PR #1, opened 2020-05-21
Adding Okhttp Client Code generation and a test harness

averabaq:
Adding Okhttp Client Code generation and a test harness


src/main/kotlin/com/cjbooms/fabrikt/configurations/PackagesConfig.kt
```diff
@@ -0,0 +1,23 @@
+package com.cjbooms.fabrikt.configurations
+
+import com.cjbooms.fabrikt.model.Destinations.clientPackage
+import com.cjbooms.fabrikt.model.Destinations.modelsPackage
+
+data class PackagesConfig(val packages: Packages) {
```
cjbooms, line 6: Looks like this could be simplified to:
```
   class Packages(val base: String) {
        val models: String  = modelsPackage(base)
        val client: String  = clientPackage(base)
    }
```

src/test/resources/examples/okHttpClient/client/LoggingInterceptor.kt
```diff
@@ -0,0 +1,25 @@
...
+import okhttp3.Interceptor
+import okhttp3.Response
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+
+class LoggingInterceptor : Interceptor {
+
+    private val logger: Logger = LoggerFactory.getLogger("GeneratedClient")
+
+    override fun intercept(chain: Interceptor.Chain): Response {
```
cjbooms, line 12: I feel like this should be a configuration option. Forcing request and response logging on the user seems like a security risk

src/main/resources/templates/client-code/logging-interceptor.kt.hbs
```diff
@@ -0,0 +1,25 @@
...
+
+class LoggingInterceptor : Interceptor {
+
+    private val logger: Logger = LoggerFactory.getLogger("GeneratedClient")
+
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val request = chain.request()
+
+        val t1 = System.nanoTime()
+        logger.info("Client Request: ${request}")
```
cjbooms, line 16: I feel like this should be a configuration option. Forcing request and response logging on the user seems like a security risk

cjbooms:
Great work @averabaq delighted to have you contributing.

src/main/resources/templates/client-code/logging-interceptor.kt.hbs
averabaq: The logging interceptor is optional already, it's always up to the user to decide whether adding this interceptor to the okhttp client at creation.

```
                httpClient = OkHttpClient.Builder()
                        .authenticator(oauth)
                        .addInterceptor(loggingInterceptor)
                        .build()
```

Issue #6, opened 2020-06-16
Automate versioning management 


averabaq:
Automate the versioning by taking the `project.version` from the latest github release

PR #8, opened 2020-06-16
Issue #6: Automate versioning management

averabaq:
Populates the project version from the git tag as described in https://github.com/palantir/gradle-git-version


cjbooms:
👍

Issue #9, opened 2020-07-29
Add support for the code generation of Spring Controllers


averabaq:
Add support for the code generation of Spring Controllers

PR #10, opened 2020-07-29
Issue #9: Add support for the code generation of Spring Controllers

averabaq:
Issue #9: Add support for the code generation of Spring Controllers

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 This library generates Kotlin code from OpenApi3 specifications to ensure your code matches your API contract. The library currently has support for:
 
 * **Jackson annotated data classes**
+* **Spring MVC annotated controllers**
+* **Service interfaces**
 * **OkHttp Client** - with the option for a resilience4j fault-tolerance wrapper
 
 ## Configuration
@@ -21,6 +23,8 @@ This section documents the available CLI parameters for controlling what gets ge
  |   `--targets`          | Targets are the parts of the application that you want to be generated.
  |                        | CHOOSE ANY OF:
  |                        |   `HTTP_MODELS` - Jackson annotated data classes to represent the schema objects defined in the input.
+ |                        |   `CONTROLLERS` - Spring annotated HTTP controllers for each of the endpoints defined in the input.
+ |                        |   `SERVICES` - Default service interfaces.
  |                        |   `CLIENT` - Simple http rest client.
 
 ## Examples
--- a/README.md
+++ b/README.md
@@ -24,7 +24,6 @@ This section documents the available CLI parameters for controlling what gets ge
  |                        | CHOOSE ANY OF:
  |                        |   `HTTP_MODELS` - Jackson annotated data classes to represent the schema objects defined in the input.
  |                        |   `CONTROLLERS` - Spring annotated HTTP controllers for each of the endpoints defined in the input.
- |                        |   `SERVICES` - Default service interfaces.
  |                        |   `CLIENT` - Simple http rest client.
 
 ## Examples
--- a/src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenOptions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenOptions.kt
@@ -13,13 +13,6 @@ enum class CodeGenerationType(val description: String, val requires: Set<Interna
             InternalCodeGenGenType.CONTROLLERS
         )
     ),
-    SERVICES(
-        "Default service implementations.",
-        setOf(
-            InternalCodeGenGenType.MODELS,
-            InternalCodeGenGenType.SERVICE_INTERFACES
-        )
-    ),
     CLIENT(
         "Simple http rest client.",
         setOf(
--- a/src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenOptions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenOptions.kt
@@ -5,6 +5,21 @@ enum class CodeGenerationType(val description: String, val requires: Set<Interna
         "Jackson annotated data classes to represent the schema objects defined in the input.",
         setOf(InternalCodeGenGenType.MODELS)
     ),
+    CONTROLLERS(
+        "Spring annotated HTTP controllers for each of the endpoints defined in the input.",
+        setOf(
+            InternalCodeGenGenType.MODELS,
+            InternalCodeGenGenType.SERVICE_INTERFACES,
+            InternalCodeGenGenType.CONTROLLERS
+        )
+    ),
+    SERVICES(
+        "Default service implementations.",
+        setOf(
+            InternalCodeGenGenType.MODELS,
+            InternalCodeGenGenType.SERVICE_INTERFACES
+        )
+    ),
     CLIENT(
         "Simple http rest client.",
         setOf(
@@ -19,6 +34,8 @@ enum class CodeGenerationType(val description: String, val requires: Set<Interna
 // Ordering is important here...
 enum class InternalCodeGenGenType {
     MODELS,
+    CONTROLLERS,
+    SERVICE_INTERFACES,
     CLIENT
 }
 
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientGenerator.kt
@@ -10,10 +10,8 @@ class OkHttpClientGenerator(
     packages: Packages,
     api: SourceApi
 ) {
-    private val simpleClientGenerator =
-        OkHttpSimpleClientGenerator(packages, api)
-    private val enhancedClientGenerator =
-        OkHttpEnhancedClientGenerator(packages, api)
+    private val simpleClientGenerator = OkHttpSimpleClientGenerator(packages, api)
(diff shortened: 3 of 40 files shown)


src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenOptions.kt
```diff
@@ -5,6 +5,21 @@ enum class CodeGenerationType(val description: String, val requires: Set<Interna
...
+    CONTROLLERS(
+        "Spring annotated HTTP controllers for each of the endpoints defined in the input.",
+        setOf(
+            InternalCodeGenGenType.MODELS,
+            InternalCodeGenGenType.SERVICE_INTERFACES,
+            InternalCodeGenGenType.CONTROLLERS
+        )
+    ),
+    SERVICES(
+        "Default service implementations.",
```
cjbooms, line 17: interfaces

README.md
```diff
@@ -21,6 +23,8 @@ This section documents the available CLI parameters for controlling what gets ge
  |   `--targets`          | Targets are the parts of the application that you want to be generated.
  |                        | CHOOSE ANY OF:
  |                        |   `HTTP_MODELS` - Jackson annotated data classes to represent the schema objects defined in the input.
+ |                        |   `CONTROLLERS` - Spring annotated HTTP controllers for each of the endpoints defined in the input.
+ |                        |   `SERVICES` - Default service interfaces.
```
cjbooms, line 27: Should this be an option? There may be some benefit to an option for generating default service implementations, with:
```
TODO("Unable to generate implementation in the absence of an 'id' parameter")
```

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientGenerator.kt
```diff
@@ -10,8 +10,10 @@ class OkHttpClientGenerator(
     packages: Packages,
     api: SourceApi
 ) {
-    private val simpleClientGenerator = OkHttpSimpleClientGenerator(packages, api)
-    private val enhancedClientGenerator = OkHttpEnhancedClientGenerator(packages, api)
+    private val simpleClientGenerator =
+        OkHttpSimpleClientGenerator(packages, api)
+    private val enhancedClientGenerator =
+        OkHttpEnhancedClientGenerator(packages, api)
```
cjbooms, line 16: is this ktlint? Looks like unnecessary line breaks

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
```diff
@@ -0,0 +1,85 @@
...
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.asTypeName
+
+object ControllerGeneratorUtils {
+    /**
+     * This maps the OpenAPI response code with a ClassName object.
+     * Pulling out the first response. This assumes first is happy path
+     * may need to revisit if we want to have conditional responses
+     */
+    fun Operation.defaultResponse(basePackage: String): Pair<Int, TypeName> {
```
cjbooms, line 22: Let's rename these functions to something more apt like:
`happyPathResponse`

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
```diff
@@ -0,0 +1,85 @@
...
+            .map { it.value?.schema }
+            .filterNotNull()
+            .firstOrNull()
+            ?.let { toModelType(basePackage, KotlinTypeInfo.from(it)) }
+            ?: Unit::class.asTypeName())
+    }
+
+    fun Operation.defaultResponseObject(): Pair<Int, Response> {
+        val toResponseMapping = responses
+            .filter { it.key != "default" }
```
cjbooms, line 37: Even more reason to rename the functions away from default, it's an overloaded term!

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
```diff
@@ -0,0 +1,85 @@
...
+            .map { (code, body) ->
+                code.toInt() to body
+            }.toMap()
+
+        // Happy path, just pull out first. Later we may have conditional responses
+        val code = responses.keys.first().toInt()
```
cjbooms, line 43: How about choosing the lowest possible HTTP code instead of the first? Seems like that simple logic might be a better happy-path

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerCodeBlocks.kt
```diff
@@ -0,0 +1,68 @@
...
+                .ok(${serviceCallLine(serviceFuncName, parameters)})
+            """
+        )
+    }
+
+    fun postWithLocationResponse(serviceFuncName: String, parameters: List<IncomingParameter>): SimpleCodeBlock {
+        return SimpleCodeBlock(
+            """val svcResp = ${serviceCallLine(serviceFuncName, parameters)}
+            val uri = %T.fromCurrentRequest().path("/${"$"}{svcResp.id}").build().toUri()
+            return ResponseEntity.created(uri).build()
```
cjbooms, line 25: Possibly more logic that should be pushed into the service layer. We are making assumptions about how people want to create their resource URIs

src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt
```diff
@@ -68,3 +97,48 @@ private fun <T : GeneratedType> groupClasses(allModels: Collection<T>): Map<T, L
...
+ */
+sealed class IncomingParameter(val oasName: String, val type: TypeName) {
+    val name: String = oasName.toKotlinParameterName()
+    fun toParameterSpecBuilder(): ParameterSpec.Builder =
+        ParameterSpec.builder(name, type)
+}
+
+class BodyParameter(oasName: String, type: TypeName, val schema: Schema) : IncomingParameter(oasName, type)
+
+class RequestParameter(
```
cjbooms, line 113: I'm a bit confused by `RequestParameter` and `BodyParameter`. `RequestParameter` has a `ParameterLocation` which could be `object BodyParam : ParameterLocation()`

README.md
averabaq: yes, I thought exactly the same. I ported it over for consistency mainly. Personally I don't find of much use getting only models and service stubs from the api spec. Should we remove it and create an issue for adding empty Service Impls?

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerCodeBlocks.kt
averabaq: Fully agreed. I guess the problem we have here is that the URI scope lays at the resource layer. How do we resolve this?

1)  Return a compose object from the service with `uri` and response object?
  ```
   data class LocationResponse<T>(T: responseObject, URI: uri)
```
2) Remove the postWithLocationResponse option in its entirety and let the user the option to construct it if he needs to?
3) Build a stub instead
4) Any other ?

File: README.md (hunk 2 above)
Comment (cjbooms) [reply README.md:27]: yes, that sounds like a good idea

File: src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientGenerator.kt (hunk 3 above)
Comment (averabaq) [reply src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientGenerator.kt:16]: I don't think so, probably some intellij auto-formatting mess up. 
will fix the format

File: src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt (hunk 6 above)
Comment (averabaq) [reply src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt:43]: absolutely :100

File: src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt (hunk 8 above)
Comment (averabaq) [reply src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt:113]: yeah, it looks like a factory pattern with the wrong inheritance. Let me have a look to see if we can easily re-factor it

File: src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerCodeBlocks.kt (hunk 7 above)
Comment (cjbooms) [reply src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerCodeBlocks.kt:25]: I'd go for 1.

File: src/test/resources/examples/customerExampleApi/controllers/Controllers.kt [hunk 9]
```diff
@@ -254,11 +253,12 @@ class DataAccessRequestsController(
...
         xFlowId: String?
-    ): ResponseEntity<Unit> {
+    ): ResponseEntity<DataAccessRequest> {
         val svcResp = service.create(dataAccessRequest, xFlowId)
-        val uri =
-            ServletUriComponentsBuilder.fromCurrentRequest().path("/${svcResp.id}").build().toUri()
-        return ResponseEntity.created(uri).build()
+        return ResponseEntity
+            .created(svcResp.first)
+            .body(svcResp.second)
```
cjbooms, line 260: is it safe to set a null body?

cjbooms:
👍

cjbooms:
👍

Issue #15, opened 2020-08-25
Wrong Service interface return type generated for GET on resource collections


cjbooms:
The following operation:
```
  /unsent-events:
    get:
      summary: "Page through all the UnsentEvent resources matching the query filters"
      tags:
        - "unsentevent"
      parameters:
        - $ref: "#/components/parameters/FlowId"
        - $ref: "#/components/parameters/TeamIdsQueryParam"
        - $ref: "#/components/parameters/AppIdsQueryParam"
        - $ref: "#/components/parameters/IncludeInactive"
        - $ref: "#/components/parameters/Limit"
        - $ref: "#/components/parameters/Cursor"
        - $ref: "#/components/parameters/TokenInfo"
      responses:
        200:
          description: "successful operation"
          headers:
            Cache-Control:
              $ref: "#/components/headers/CacheControl"
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/UnsentEventQueryResult"
        default:
          description: "error occurred - see status code and problem object for more\
            \ information."
          content:
            application/problem+json:
              schema:
                $ref: "https://zalando.github.io/problem/schema.yaml#/Problem"
      security:
        - oauth2:
            - "uid"
            - "fabric-event-scheduler.read"
```

is producing a Service interface with the following signature:
```
  fun query(
    limit: Int,
    xTokeninfoForward: TokenInfo,
    xFlowId: String?,
    teamIds: List<String>?,
    appIds: List<String>?,
    includeInactive: Boolean?,
    cursor: String?
  ): List<UnsentEvent>
```

But it should be producing this:
```
  fun query(
    limit: Int,
    xTokeninfoForward: TokenInfo,
    xFlowId: String?,
    teamIds: List<String>?,
    appIds: List<String>?,
    includeInactive: Boolean?,
    cursor: String?
  ): UnsentEventQueryResult
```

PR #16, opened 2020-08-26
Issue #15 Fixing return type on Service interfaces 

cjbooms:
It will now match the expected return type in generated controllers for query operations. 

Also adding in spring mvc as a test dependency so compilations issues are visible in the example resources


cjbooms:
👍

PR #29, opened 2020-12-14
ISSUE#23 Ensures enums are only generated for x-extensible-enum when enabled

### LINKED ISSUE #23: Add a configuration option to treat x-extensible-enum as an enum
Zalando's string format `x-extensible-enum` could be generated as an enum if requested by users.

The default should be not to treat these strings as enums as it could cause clients to break as new enum values are added. But for server-side development, it may be nicer to generate these definitions as enums

cjbooms:
Closes #23


src/main/kotlin/com/cjbooms/fabrikt/generators/MutableSettings.kt
```diff
@@ -0,0 +1,20 @@
...
+
+import com.cjbooms.fabrikt.cli.ClientCodeGenOptionType
+import com.cjbooms.fabrikt.cli.CodeGenerationType
+import com.cjbooms.fabrikt.cli.ModelCodeGenOptionType
+
+object MutableSettings {
+    var generationTypes: MutableSet<CodeGenerationType> = mutableSetOf()
+    var modelOptions: MutableSet<ModelCodeGenOptionType> = mutableSetOf()
+    var clientOptions: MutableSet<ClientCodeGenOptionType> = mutableSetOf()
+    fun updateSettings(
```
cjbooms, line 11: Feels quite wrong doing this, but it makes life so much easier, and given the context it's fine

averabaq: It makes a lot of sense to me 👍

averabaq:
👍

PR #31, opened 2020-12-15
ISSUE#25 When generating inlined object definitions prefix derived Class names 

### LINKED ISSUE #25: Inlined object definitions should be Prefixed with their parent object name
Inlined objects that are generated often share the same name as top level objects, or other inlined objects. This causes models to overwrite each other at generation time with unexpected results
```
openapi: 3.0.0
components:
  schemas:
    ContainsInLinedObject:
      type: object
      properties:
        generation:
          type: object
          properties:
            call_home:
              type: object
              required:
                - url
              properties:
                url:
                  type: string
            database_view:
              type: object
              required:
                - view_name
              properties:
                view_name:
                  type: string
            direct:
              type: string
```
Should produce:
```
data class ContainsInLinedObject(
    @param:JsonProperty("generation")
    @get:JsonProperty("generation")
    @get:Valid
    val generation: ContainsInLinedObjectGeneration? = null
)

data class ContainsInLinedObjectGenerationCallHome(
    @param:JsonProperty("url")
    @get:JsonProperty("url")
    @get:NotNull
    val url: String
)

data class ContainsInLinedObjectGenerationDatabaseView(
    @param:JsonProperty("view_name")
    @get:JsonProperty("view_name")
    @get:NotNull
    val viewName: String
)

data class ContainsInLinedObjectGeneration(
    @param:JsonProperty("call_home")
    @get:JsonProperty("call_home")
    @get:Valid
    val callHome: ContainsInLinedObjectGenerationCallHome? = null,
    @param:JsonProperty("database_view")
    @get:JsonProperty("database_view")
    @get:Valid
    val databaseView: ContainsInLinedObjectGenerationDatabaseView? = null,
    @param:JsonProperty("direct")
    @get:JsonProperty("direct")
    val direct: String? = null
)



```

cjbooms:
Closes #25 

When FabriKt encounters an inlined object definition it derives a class name using the property key. This was brittle and caused collisions for common property keys such as `type` or `name` that contained inline schemas.

To help alleviate the problem FabriKt now prefixes the derived class name with the Class Name of the enclosing top level schema definition. 

This will stop collisions between separate Schema entries which is the common case. Collisions will still occur if a single schema definition contains multiple inlined schema definitions, at different depths, that share key names.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
@@ -29,8 +29,10 @@ import com.cjbooms.fabrikt.util.KaizenParserExtensions.isReferenceObjectDefiniti
 import com.cjbooms.fabrikt.util.KaizenParserExtensions.mappingKey
 import com.cjbooms.fabrikt.util.KaizenParserExtensions.safeName
 import com.cjbooms.fabrikt.util.KaizenParserExtensions.toMapValueClassName
+import com.cjbooms.fabrikt.util.KaizenParserExtensions.toModelClassName
 import com.cjbooms.fabrikt.util.NormalisedString.toModelClassName
 import com.reprezen.kaizen.oasparser.model3.Discriminator
+import com.reprezen.kaizen.oasparser.model3.Schema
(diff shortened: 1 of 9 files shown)


averabaq:
👍

PR #35, opened 2021-01-26
Improve external reference support

cjbooms:
Fixes the issue where generation stops when an array element in an externally referenced schema is encountered


averabaq:
👍

cjbooms:
👍

Issue #30, opened 2020-12-15
Generate Controller Interfaces instead of implementations


cjbooms:
Code generation of Controllers would be greatly simplified if we focused on generating interfaces rather than implementations. We would no longer need to generate service interfaces

PR #37, opened 2021-04-28
Issue #30: Generate Controller Interfaces instead of implementations

averabaq:
Issue #30: Generate Controller Interfaces instead of implementations


cjbooms:
👍

PR #39, opened 2021-04-30
Issue #34: Optional properties with a default should not be nullable

averabaq:
Issue #34: Optional properties with a default should not be nullable


cjbooms:
👍

PR #40, opened 2021-05-01
Fixing bug with constructor when multiple objects under an allOf

cjbooms:
Fixing bug with constructor when multiple objects under an allOf, but only one is the super class


cjbooms:
👍

PR #42, opened 2021-05-01
These static imports are important, they were accidentally removed 

cjbooms:
When transitioning from Controller implementations to interfaces. There is a hole in our tests that doesn't pick this up, but it would be a time-consuming test to rejig


cjbooms:
👍

PR #43, opened 2021-05-02
Restoring ResponseEntity wrapper 

cjbooms:
Controller return types were accidentally changed in the migration to interfaces, and tests were adjusted to verify the bad behaviour.


cjbooms:
👍

PR #45, opened 2021-05-14
ISSUE#44 Fixing client generation code for generic typed responses 

cjbooms:
Return types such as `Map<String, Any>?` are now supported using TypeReference


cjbooms:
👍

PR #50, opened 2021-06-03
Improving client

cjbooms:
- Making the OAuth class a general purpose Interceptor and Authenticator that takes a token function
- Improving formatting of client classes


cjbooms:
👍

PR #52, opened 2021-06-08
Fixing bug when generating HTTP parameters on an non-exploded format.

lagarino:
The parameters are concatenated by default with ", ", leaving a space between the comma and the next parameter


cjbooms:
Thanks @lagarino ! 
Released now.

PR #54, opened 2021-06-17
Improvements to class generation for complex polymorphism

cjbooms:
- Support for multiple discriminator keys pointing to the same class
 - Preventing the top-level use of oneOf anyOf or allOf in a schema of type object as OpenApi3 is vague on how to interpret this. Providing a good warning message when this happens


cjbooms:
👍

PR #55, opened 2021-06-18
Improvements to inlined enum class generation.

cjbooms:
Inlined enums now get a class name prefixed with the class name that declares them. This should help avoid conflicts between enums with the same name inlined in different schemas


cjbooms:
👍

averabaq:
👌  🙂

PR #61, opened 2021-07-08
ISSUE-60 Adding logic to detect when an schema is referenced in an external API document

cjbooms:
When this occurs, simply parse the entire external API spec, and generate models for every schema within it. We could be more targeted with this but attempts I made to do so were resulting in quite complicated code.


cjbooms:
👍

src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
```diff
@@ -114,23 +118,34 @@ class JacksonModelGenerator(
...
-                }.toMutableSet()
+    private val primaryDocUrl = sourceApi.openApi3.getDocumentUrl()
+    private val externallyReferencedApis = mutableSetOf<String>()
+
+    private fun <V> IJsonOverlay<V>.getDocumentUrl() = Overlay.of(this).positionInfo.get().documentUrl
 
+    fun generate(): Models {
+        val models: MutableSet<TypeSpec> = createModels(sourceApi.openApi3, sourceApi.allSchemas)
+        externallyReferencedApis.forEach { docUrl ->
+            val api = OpenApi3Parser().parse(URL(docUrl))
```
averabaq, line 129: awesome! 🤩

averabaq:
Sorry, I've just seen it! Awesome 😃

PR #72, opened 2021-07-09
ISSUE-60 More targeted capturing of external schemas.

cjbooms:
Don't attempt to generate external schemas from current API document. Save generation for external document processing


cjbooms:
👍

PR #74, opened 2021-08-17
Fix typed map issue

cjbooms:
Fixing an issue where complex nested maps definitions failed to generate correctly when referencing other typed objects in the OAS scheama. Test case expanded to cover this scenario


averabaq:
👍

PR #77, opened 2021-08-18
Fixing NPE experienced when testing 1.3.0 

cjbooms:
Encountered a Null Pointer Exception when testing 1.3.0 from the Sonatype staging repo in some Zalando projects


cjbooms:
👍

PR #83, opened 2021-10-11
Adding the ability to set arbitrary additional HTTP headers. 

cjbooms:
This will be convenient for things like propagating Open-Tracing telemetry headers


averabaq:
👍

PR #87, opened 2021-11-25
Adding reverse lookup function to generated enums

cjbooms:
Adds the following `fromValue` reverse lookup function to generated enums:
```kotlin
enum class ContentType(
    @JsonValue
    val value: String
) {
    APPLICATION_JSON("application/json"),

    APPLICATION_X_SOME_TYPE_JSON("application/x.some-type+json"),

    APPLICATION_X_SOME_OTHER_TYPE_JSON_VERSION_2("application/x.some-other-type+json;version=2");

    companion object {
        private val mapping: Map<String, ContentType> = values().associateBy(ContentType::value)

        fun fromValue(value: String): ContentType? = mapping[value]
    }
}
```


cjbooms:
👍

cjbooms:
Hope this is ok @averabaq I need it right now in a Zalando project

PR #90, opened 2021-12-02
Multiple response content types

herojan:
Currently if there are multiple response schemas, the client just generates the response class for the first one. This changes it so that it generates JsonNode when there are multiple possible response schemas.


src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
```diff
@@ -16,18 +18,24 @@ object ClientGeneratorUtils {
...
      */
     fun Operation.toClientReturnType(packages: Packages): TypeName {
-        val returnType = this.getPrimaryContentMediaType()?.let {
-            toModelType(
-                packages.base,
-                KotlinTypeInfo.from(it.value.schema)
-            )
-        } ?: Unit::class.asTypeName()
+        val returnType =
+                if (hasMultipleResponseSchemas()) {
```
cjbooms, line 28: Does this include the default Problem json, or is it just types outside of that?

herojan: Ah it does, good catch. Fixed

PR #95, opened 2021-12-06
Fix enum reference

herojan:
The reference to the enum value `ParentType.CHILD_TYPE_A`  in `ChildTypeA` in the example, was previously referencing `ParentType.CHILDTYPEA`


cjbooms:
My bad, I broke this a couple of weeks ago when I improved the enum formatting ability for special characters. Will cut a bug fix release now.

PR #97, opened 2021-12-06
Use x-extensible-enum value for json subtype annotation

herojan:
When the discriminator is an x-extensible-enum, the enum value should be used to select the subtype, not the class name as a string.


cjbooms:
👍

PR #105, opened 2022-03-02
Fix inline generation bug #103

### LINKED ISSUE #103: sealed class generation breaks when referenced from another schema
It looks like polymorphic class generation breaks as soon as that class is referenced from another schema. Here's an example that adds a `Wrapper` schema to the yml from the readme:

```diff
openapi: 3.0.0
components:
  schemas:
+    Wrapper:
+      type: object
+      properties:
+        polymorph:
+          $ref: '#/components/schemas/PolymorphicEnumDiscriminator'
    PolymorphicEnumDiscriminator:
      type: object
      discriminator:
        propertyName: some_enum
        mapping:
          obj_one: '#/components/schemas/ConcreteImplOne'
          obj_two: '#/components/schemas/ConcreteImplTwo'
      properties:
        some_enum:
          $ref: '#/components/schemas/EnumDiscriminator'
    ConcreteImplOne:
      allOf:
        - $ref: '#/components/schemas/PolymorphicEnumDiscriminator'
        - type: object
          properties:
            some_prop:
              type: string
    ConcreteImplTwo:
      allOf:
        - $ref: '#/components/schemas/PolymorphicEnumDiscriminator'
        - type: object
          properties:
            some_prop:
              type: string
    EnumDiscriminator:
      type: string
      enum:
        - obj_one
        - obj_two
```

In the generated code, `PolymorphicEnumDiscriminator` is generated as a regular `class` and the subtypes are missing:

```kotlin
data class Wrapper(
  @param:JsonProperty("polymorph")
  @get:JsonProperty("polymorph")
  @get:Valid
  val polymorph: PolymorphicEnumDiscriminator? = null
)

class PolymorphicEnumDiscriminator() {
  @param:JsonProperty("some_enum")
  @get:JsonProperty("some_enum")
  @get:NotNull
  val someEnum: EnumDiscriminator
}

enum class EnumDiscriminator(
  @JsonValue
  val value: String
) {
  OBJ_ONE("obj_one"),

  OBJ_TWO("obj_two");

  companion object {
    private val mapping: Map<String, EnumDiscriminator> =
        values().associateBy(EnumDiscriminator::value)

    fun fromValue(value: String): EnumDiscriminator? = mapping[value]
  }
}
```

cjbooms:
A bug was introduced that was causing top-level schema definitions to be generated as part of the inlined generation logic. This resulted in complex polymorphic classes getting over-written with simpler inline-derived versions.

This PR removes the logic that generates referenced schemas during inline generation. It relies on the fact that these schemas will undergo regular/full generation.

Closes #103

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
@@ -199,15 +199,7 @@ class JacksonModelGenerator(
                     val inlinedModels = buildInLinedModels(props, enclosingSchema, apiDocUrl)
                     inlinedModels + currentModel
                 }
-                is PropertyInfo.ObjectRefField ->
-                    when {
-                        it.schema.isReferenceObjectDefinition() ->
-                            it.schema.topLevelProperties(HTTP_SETTINGS, enclosingSchema).let { props ->
-                                buildInLinedModels(props, enclosingSchema, apiDocUrl) +
-                                    standardDataClass(it.schema.safeName().toModelClassName(), props)
-                            }
-                        else -> emptySet()
-                    }
+                is PropertyInfo.ObjectRefField -> emptySet() // Not an inlined definition, so do nothing
                 is PropertyInfo.MapField ->
                     buildMapModel(it)?.let { mapModel -> setOf(mapModel) } ?: emptySet()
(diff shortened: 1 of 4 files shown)


cjbooms:
👍

cjbooms:
And by `A bug was introduced`, I off-course mean: `I introduced a bug`

I had so many different iterations of trying to get generation from external files working I obviously broke something

averabaq:
👍

PR #115, opened 2022-03-07
ISSUE-111: Support naming controller methods after the operationId

### LINKED ISSUE #111: Support naming controller methods after the operationId
In openapi-generator, controller methods are named after the `operationId` in the spec file. fabrikt just seems to name the methods after the HTTP method (`post()` etc). This makes it impossible to implement multiple controllers with overlapping HTTP methods in the same class.

The simplest solution would be to use the `operationId` like openapi-generator does - perhaps as an option as it would be breaking otherwise?

I could also see wanting to have this feature even apart from the mentioned multiple-controllers-in-one-class use case, simply because the `operationId`, if present, is probably carries more information than the generated one.

averabaq:
Controller methods are now named after the operationId if present and fallback to http verb otherwise

Closes #111

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
@@ -78,6 +78,9 @@ object ControllerGeneratorUtils {
 
     fun controllerName(resourceName: String) = "$resourceName${ControllerType.SUFFIX}"
 
-    fun methodName(verb: String, isSingleResource: Boolean) =
+    fun methodName(op: Operation, verb: String, isSingleResource: Boolean) =
+        op.operationId ?: httpVerbMethodName(verb, isSingleResource)
+
+    private fun httpVerbMethodName(verb: String, isSingleResource: Boolean) =
         if (isSingleResource) "${verb}ById" else verb
 }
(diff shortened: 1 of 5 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
```diff
@@ -78,6 +78,9 @@ object ControllerGeneratorUtils {
 
     fun controllerName(resourceName: String) = "$resourceName${ControllerType.SUFFIX}"
 
-    fun methodName(verb: String, isSingleResource: Boolean) =
+    fun methodName(op: Operation, verb: String, isSingleResource: Boolean) =
+        op.operationId ?: httpVerbMethodName(verb, isSingleResource)
```
averabaq, line 82: The `operationId` is treated as-is. Perhaps we should manipulate it by some sort of normalization, uncapitalize first letter, camel case,... Any thought?

cjbooms: Agreed. Stick something in `NormalisedString` class or reuse `camelToSnake()`

src/test/kotlin/com/cjbooms/fabrikt/generators/SpringControllerGeneratorTest.kt
```diff
@@ -140,6 +139,21 @@ class SpringControllerGeneratorTest {
         ).isTrue()
     }
 
+    @Test
+    fun `ensure controller methods has the correct method names`() {
```
cjbooms, line 143: Looks good to me, but I would say just piggy-back on one of the existing tests instead of duplicating the resources

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/ControllerGeneratorUtils.kt
averabaq: 👍

averabaq: Going for `NormalisedString.camelCase()` as an attempt to be pragmatic here

cjbooms:
👍

PR #116, opened 2022-03-07
Treat simple oneOf as Any 

### LINKED ISSUE #110: AnyType for oneOf
This is a followup for #106.

openapi-generator considers `foo` to be of type "any"  in the following example using `oneOf`:

```yml
openapi: 3.0.0
components:
  schemas:
    Foo:
      type: object
      properties:
        foo:
          oneOf:
            - type: string
            - type: object
              properties:
                a:
                  type: string
                b:
                  type: string

```

It looks like right now fabrikt just uses the first schema here:

```kotlin
data class Foo(
  @param:JsonProperty("foo")
  @get:JsonProperty("foo")
  val foo: String? = null
)
```

cjbooms:
A oneOf that can not be resolved to a polymorphic definition should be treated as an Any.

Closes #110

Also fixes the duplicate key handling by replacing a `results.toSet()` with `results.distinctBy { it.oasKey }`


src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -162,17 +162,7 @@ sealed class PropertyInfo {
         val maximum: Number? = schema.safeField(Schema::getMaximum)
         val exclusiveMaximum: Boolean? = schema.safeField(Schema::getExclusiveMaximum)
 
-        private fun <T> Schema.safeField(getField: Schema.() -> T?): T? =
```
cjbooms, line 165: I have no idea what this was trying to do.

cjbooms:
👍

PR #57, opened 2021-06-23
Add support for the UUID format of strings

cjbooms:
Closes #56

PR #118, opened 2022-03-11
Map strings with format: uri to java.net.URI

Gama11:
Similar to the UUID handling that has already been added (#57). This matches the behavior of openapi-generator.


Gama11:
No, take your time. :)

cjbooms:
Released under 4.0.0 as it's a breaking change

PR #122, opened 2022-03-14
Force type to be object if any properties are defined

### LINKED ISSUE #121: allOf with single $ref leads to Any
Code generation for the following example is correct (return types correctly typed as `Result`):

```yml
openapi: 3.0.0
paths:
  /test:
    get:
      operationId: test
      responses:
        '200':
          description: Operation successful
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Result'
components:
  schemas:
    Result:
      allOf:
        - $ref: '#/components/schemas/Base'
        - type: object
          properties:
            bar:
              type: string
    Base:
      properties:
        foo:
          type: string
```

However, when `Result` is reduced to just this, the return type in both the client and the controller is suddenly `Any`:

```yml
    Result:
      allOf:
        - $ref: '#/components/schemas/Base'
```

I think an `allOf` with a single item should be treated the same as a direct `$ref`.

cjbooms:
Force type to be object if any properties are defined
Closes #121

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/model/SourceApi.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/model/SourceApi.kt
@@ -8,6 +8,7 @@ import com.reprezen.kaizen.oasparser.model3.OpenApi3
 import com.reprezen.kaizen.oasparser.model3.Schema
 import java.nio.file.Path
 import java.nio.file.Paths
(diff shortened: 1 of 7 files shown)


Gama11:
Perhaps a warning should also be printed? Personally I'd still prefer to have an explicit `type: object` in my specs, a warning would allow to easily find and fix these cases.

PR #124, opened 2022-03-16
Non-Nullable client return types

### LINKED ISSUE #120: Client return type is always nullable
I noticed that the return type of clients is always nullable:

https://github.com/cjbooms/fabrikt/blob/fe780487196eb641eafac706d20b822098eae99b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt#L40

Is this simply a result of OkHttp's `response.body` being nullable? For endpoints that return something, it might be more sensible to throw an `ApiException` instead if the status code was 2xx but there was no body. Otherwise user code always needs a try-catch _and_ null handling, which seems annoying.

averabaq:
ISSUE-120: The http client will now return a non-nullable response type ~~on those operations that define a unique 200 successful response~~ such as follows:

```
no content -> ApiResponse<Unit>
single content schema: MyTypeOne -> ApiResponse<MyTypeOne>
multiple content schemas: -> ApiResponse<JsonNode> or ApiResponse<Any> 
```

Closes #120

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
@@ -46,8 +46,7 @@ object ClientGeneratorUtils {
                     } ?: Unit::class.asTypeName()
                 }
 
-        return "ApiResponse".toClassName(packages.client)
-            .parameterizedBy(returnType.copy(nullable = true))
+        return "ApiResponse".toClassName(packages.client).parameterizedBy(returnType)
     }
 
     fun simpleClientName(resourceName: String) = "$resourceName${ClientType.SIMPLE_CLIENT_SUFFIX}"
--- a/src/main/resources/templates/client-code/http-util.kt.hbs
+++ b/src/main/resources/templates/client-code/http-util.kt.hbs
@@ -34,7 +34,7 @@ fun Headers.Builder.header(key: String, value: String?): Headers.Builder = this.
 }
 
 @Throws(ApiException::class)
-fun <T> Request.execute(client: OkHttpClient, objectMapper: ObjectMapper, typeRef: TypeReference<T>): ApiResponse<T?> =
+fun <T> Request.execute(client: OkHttpClient, objectMapper: ObjectMapper, typeRef: TypeReference<T>): ApiResponse<T> =
     client.newCall(this).execute().use { response ->
         when {
             response.isSuccessful ->
--- a/src/test/resources/examples/okHttpClient/client/ApiClient.kt
+++ b/src/test/resources/examples/okHttpClient/client/ApiClient.kt
@@ -70,7 +70,7 @@ class ExamplePath1Client(
         content: Content,
         explodeListQueryParam: List<String>?,
         additionalHeaders: Map<String, String> = emptyMap()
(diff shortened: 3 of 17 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
```diff
@@ -37,10 +38,12 @@ object ClientGeneratorUtils {
...
         return "ApiResponse".toClassName(packages.client)
-            .parameterizedBy(returnType.copy(nullable = true))
+            .parameterizedBy(returnType.copy(nullable = isNullableReturnType()))
     }
 
     fun simpleClientName(resourceName: String) = "$resourceName${ClientType.SIMPLE_CLIENT_SUFFIX}"
 
     fun enhancedClientName(resourceName: String) = "$resourceName${ClientType.ENHANCED_CLIENT_SUFFIX}"
+
+    private fun Operation.isNullableReturnType(): Boolean = responses.all { it.key == "200" }.not()
```
cjbooms, line 48: This seems a bit brittle. Is there nothing in theAPI spec that gives an indication whether the response should be nullable or not?

averabaq: Not sure 😬  I guess the right approach would be to always to rely on the `content` attribute of the response. If present then we may always treat it as `non-nullable`, `nullable` otherwise.

According to the spec:
> Some responses, such as 204 No Content, have no body. To indicate the response body is empty, do not specify a content for the response

The thing is that if the `content` attribute isn't specified then we can't know the return type, so making it nullable or non-nullable is kind of irrelevant.

Not sure what to do here. Perhaps we should always treat them as non-nullable? 🤔

averabaq: Cases like the one below look "semantically" wrong, but still "syntactically" correct. I guess the question in here is if there could exist any scenario where the response body could be optional?

```
      responses:
        201:
          description: "successful operation"
          headers:
            Location:
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/NullableObject'
```

cjbooms: Yes, in that case I guess `NullableObject` is an object containing nullable properties. It itself is non nullable.

> Some responses, such as 204 No Content, have no body. To indicate the response body is empty, do not specify a content for the response

Do we currently have any use for nullable?  I'm thinking the following:
no content -> `ApiResponse<Unit>`
single content schema: MyTypeOne -> `ApiResponse<MyTypeOne>`
multiple content schemas:  -> `ApiResponse<JsonNode>` or `ApiResponse<Any>`

So I think making all response types non-nullable is the right thing to do.

averabaq: That makes perfect sense ☝️ 

> Do we currently have any use for nullable?

None on the top of my head right now except for Streaming APIs upon empty streams, which are not the case. I can't think of anything else

This `non-nullable` approach is much more consistent and beautiful 😍 

PR update coming.... 👍

PR #126, opened 2022-03-18
fix broken codegen for put/post/patch requests without request body

pschichtel:
The `.` got lost. Also okhttp3 always once *some* payload, so in cases where no data should be sent, an empty buffer must be passed in.


cjbooms:
👍

PR #127, opened 2022-03-21
Add Support for Path Level Parameters and avoid Name Clashes

pschichtel:
This PR does 3 things actually:

1. support path level parameters (including the override mechanic)
2. adds bodies to generated kdocs
3. renames all parameters if a name conflict is detected by prefixing all parameters with the location (body for request bodies)

Compatibility:

In some places the parameter order changed which is probably breaking.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/GeneratorUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/GeneratorUtils.kt
@@ -1,6 +1,5 @@
 package com.cjbooms.fabrikt.generators
 
-import com.cjbooms.fabrikt.generators.controller.ControllerGeneratorUtils
 import com.cjbooms.fabrikt.generators.model.JacksonModelGenerator.Companion.toModelType
 import com.cjbooms.fabrikt.model.BodyParameter
 import com.cjbooms.fabrikt.model.IncomingParameter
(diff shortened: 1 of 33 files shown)


pschichtel:
the "additionalHeaders" parameter is not yet part of the name clash avoidance, I think a generic way to generate parameter default values would be needed first. It's definitely desirable, since it would also get rid of a few special cases.

cjbooms:
Nice work

> the "additionalHeaders" parameter is not yet part of the name clash avoidance, I think a generic way to generate parameter default values would be needed first. It's definitely desirable, since it would also get rid of a few special cases.

I think that's fine. We can address that in another PR if anyone hits the issue.

> In some places the parameter order changed which is probably breaking.
That's ok, we have another PR merged that is also a breaking change. We can rev a major version for both

src/main/kotlin/com/cjbooms/fabrikt/generators/GeneratorUtils.kt
```diff
@@ -134,4 +142,70 @@ object GeneratorUtils {
...
+                    it.description,
+                    toModelType(basePackage, KotlinTypeInfo.from(it.schema), isNullable(it)),
+                    it
+                )
+            }
+            .sortedBy { it.type.isNullable }
+
+        val all = bodies + parameters + extraParameters
+
+        return if (all.map { it.name }.toSet().size != all.size) {
```
cjbooms, line 180: Maybe extract this logic to a new function to avoid or reduce the necessary code comments.
eg, something like:
```
val allParameters = bodies + parameters + extraParameters
val distinctParameters = maybeRenameParametersToAvoidCoillisions(all)
return distinctParameters
```

PR #143, opened 2022-08-03
Remove LoggingInterceptor.

cjbooms:
We should not be generating a Logging interceptor that is pinned to specific logging implementation. Removing this so people can choose whatever logging framework they like


cjbooms:
👍

PR #152, opened 2022-11-08
Move additional properties map to data class constructor.

cjbooms:
This will ensure that additional properties are taken into account with equality check. Previously data classes containing different additional properties were being reported as equal if all constructor parameters matched


cjbooms:
👍

PR #155, opened 2022-12-29
Add OpenApi 3.1 schema example.

cjbooms:
Adds an example OpenApi 3.1 spec as taken from:
https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.1/webhook-example.yaml


averabaq:
👍

PR #160, opened 2023-01-24
Include a model generator option to generate a companion object

### LINKED ISSUE #158: Adding a companion object to generated data classes
Hey team,
I was wondering if this request would be considered reasonable. Currently while generating model classes, they are realized as data classes. While this already has benefits over regular Kotlin classes, it can be extended further. Given that Kotlin has the ability to add extension functions, the current realisation only allows for extension functions at the instance level. To enable fluent semantics, it would be an option to accommodate extension functions at the class level through the generation of a companion object.

The request is illustrated as below:
Source spec: https://raw.githubusercontent.com/OAI/OpenAPI-Specification/main/examples/v3.0/petstore.yaml
Generated Pet model class:
```
package com.example.models
... imports

data class Pet(
... properties
)
```

The usage of this generated model class:
```
// regular code flow
return Pet(... values)
```

Generated model class after incorporating proposal:
```
package com.example.models
... imports

data class Pet(
... properties
)  {
  companion object
}
```

The enablement of class level extension functions:
```
fun Pet.Companion.from(... values) {
    return Pet(... values)
}

// regular code flow
return Pet.from(... values)
```

Please let me know what you think about this. I have a patch where I attempted this already and it works. I'll polish it and put out a PR in case I get a green signal.

rajki:
Closes #158


cjbooms:
👍

cjbooms:
Nice work. This looks good to me. Is there anything else you wanted to add or is this ready for merge?

rajki:
Thanks for the quick review! I think we are good to merge.

Issue #154, opened 2022-12-15
Support Micronaut server generation


atollk:
For the project I am a part of, we are using the Micronaut framework to define our HTTP servers. We are looking to switch from hand-written server classes to classes generated by an OpenAPI spec. We'd like to use fabrikt for several reasons, one of them being that the code is much easier to understand compared to the template files used by the "official" OpenAPI generator.

Our idea is to provide the initial work to add Micronaut support to Fabrikt (which should be rather easy, considering it is quite similar to Spring) ourselves, to then benefit from any future updates to the general Fabrikt code base too. Would that be something you'd welcome as PRs or would you rather keep Fabrikt focused on just Spring?

PR #161, opened 2023-02-11
Support Micronaut controllers

atollk:
Fixes https://github.com/cjbooms/fabrikt/issues/154

At the moment, this is yet a draft. I haven't done any tests (neither automated unit tests, nor manual tests in practice) so far and might need to hash out some details for code generation. However, I would appreciate some input already, whether this is a solution which you'd be okay with in general.


cjbooms:
Hi @atollk and thanks for your contribution. Apologies for the delayed response, I was on vacation until yesterday.

I like this approach and would be happy to accept the contribution. One question, did you consider just defining a new target instead of adding a tertiary controller configuration option? 
`CodeGenerationType.MICRONAUT_CONTROLLERS`

atollk:
Hey, no worries. I didn't consider that option because it would feel somewhat odd to me that "models" and "Spring controllers" (which you'll often generate together) have the same relation to each other as "Micronaut controllers" and "Spring controllers" (which you'd probably need both at once). But I don't see any hard problems with that design either. So if you want me to change that, I can do it no problem.

Otherwise, I'll probably try to add some additional tests to this change and come back to you when I consider it completely ready for review.

cjbooms:
> Hey, no worries. I didn't consider that option because it would feel somewhat odd to me that "models" and "Spring controllers" (which you'll often generate together) have the same relation to each other as "Micronaut controllers" and "Spring controllers" (which you'd probably need both at once). But I don't see any hard problems with that design either. So if you want me to change that, I can do it no problem.
> 
> Otherwise, I'll probably try to add some additional tests to this change and come back to you when I consider it completely ready for review.

Sounds good. Happy to go with your original approach, it'll work well.

cjbooms:
> Hey, no worries. I didn't consider that option because it would feel somewhat odd to me that "models" and "Spring controllers" (which you'll often generate together) have the same relation to each other as "Micronaut controllers" and "Spring controllers" (which you'd probably need both at once). But I don't see any hard problems with that design either. So if you want me to change that, I can do it no problem.
> 
> Otherwise, I'll probably try to add some additional tests to this change and come back to you when I consider it completely ready for review.

Sounds good. Happy to go with your original approach, it'll work well.

cjbooms:
Moved to #164

PR #180, opened 2023-03-28
Support jakarta.validation for model classes

zapodot:
Added CLI option `--validation-library` with default value JAVAX_VALIDATION and optional value JAKARTA_VALIDATION to allow generated model classes to support either


cjbooms:
Nice work. Given the default behaviour remains unchanged, I'll release this a minor version

Issue #89, opened 2021-12-02
support kotlinx.serialization


christophsturm:
This library looks great, are you planning to add support for kotlinx serialization, or would you accept a pr that adds it?

cjbooms:
I would be very receptive to PRs, but have no experience with kotlinx serialization.

ulrikandersen:
@zettelmj Have you been able to make progress on adding Kotlinx Serialization to the library?

Serisium:
@zettelmj and @ulrikandersen, I am interested in tackling this feature. How has your experience been so far, and do you have any advice for getting started?

Thank you for your work so far!

ulrikandersen:
@Serisium I am happy to help with code reviews and feedback along the way! 

@zettelmj has taken the first steps in #249, so I guess that would be the place to start.

zettelmj:
Hi Sorry for the late answer. I did try to tackle the whole problem at once. In retrospective that is most probably a step too big. I should have gone for the model part first. I would suggest the same: Add the code that allows for the model part of the process in first. I am planning to look into adding end 2 end tests for the existing Jackson based serialisation over the weekend.

dshatz:
Hi!

I decided to try implementing a Ktor-client support and stumbled upon this issue with kotlinx-serialization models.
Here is the openapi file: https://github.com/cjbooms/fabrikt/commit/e312a8e654dd856fc3b8ec59b01d7113ba3bc7c3
I get this `Content.kt` when generating with `end2end-tests:models-kotlinx:generateCode`:

```kotlin
package com.example.models

import com.fasterxml.jackson.`annotation`.JsonSubTypes
import com.fasterxml.jackson.`annotation`.JsonTypeInfo
import kotlin.String
import kotlinx.datetime.Instant

@JsonTypeInfo(
  use = JsonTypeInfo.Id.NAME,
  include = JsonTypeInfo.As.EXISTING_PROPERTY,
  property = "model_type",
  visible = true,
)
@JsonSubTypes(JsonSubTypes.Type(value = FirstModel::class, name =
    "first_model"),JsonSubTypes.Type(value = SecondModel::class, name =
    "second_model"),JsonSubTypes.Type(value = ThirdModel::class, name = "third_model"))
public sealed class Content(
  public open val id: String? = null,
  public open val firstAttr: Instant? = null,
  public open val secondAttr: String? = null,
  public open val thirdAttr: ContentThirdAttr? = null,
  public open val etag: String? = null,
[...]

PR #181, opened 2023-04-01
General dependency update

chris-hatton:
I am interested in taking on issue #89, to support `kotlinx.serialization`.

In preparation, I checked out `master` and observed that dependencies were generally somewhat out-of-date.  I've updated them here before adding support for the latest `kotlinx.serialization` version.

A couple of minor changes arose to account for the new dependencies:

1. Dokka now prefers to expose a task-per-output-format, than have a single output format selected.  I imagine this is to support output of multiple formats.  The way the task is acquired for the `kotlinDocJar` task then had to change.
2. Kotlin Poet now supports multple 'simpleName's; I'm not sure why to be honest but fixing our usage was a simple case of removing the single argument label.

Unit tests passing locally ✅

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,10 +2,10 @@ import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import org.jetbrains.dokka.gradle.DokkaTask
 
 plugins {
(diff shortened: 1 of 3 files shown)


cjbooms:
👍

cjbooms:
Thanks @chris-hatton

PR #186, opened 2023-04-06
Add authentication to controller generators

### LINKED ISSUE #162: Controllers: Add more flexibility on how to extract information from requests
At the moment, query parameters, headers, and request bodies are extracted into specific variables with their respective annotations; nothing more, nothing less. Spring (and Micronaut) also supports storing the entire request information in one single object. There is also an `Authentication` class to provide information about the authenticated user. Both of these can be retreived by also adding them to the parameter list of a controller function with their proper annotations.

I feel like it would be helpful to have additional controller options for things like "enable complete request", "enable authentication info", and "disable single variables".

ttobollik:
Partly fixes #162 
Added authentication to controller generation.  
Added test cases. 
Running ktlint led to additional modifications which don't regard the authentication use case.
Current state does not add a CLI flag for conditional generation of this feature, contrary to the original description of the feature.


cjbooms:
We need a flag to allow builders to opt-in to this code generation feature. For companies that perform authentication in edge services, their is no need for these code-level security constructs

src/test/kotlin/com/cjbooms/fabrikt/generators/MicronautAuthenticationTest.kt
```diff
@@ -0,0 +1,203 @@
+package com.cjbooms.fabrikt.generators
```
cjbooms, line 1: It would be good to see these tests either replaced with or complemented with the existing style of test where code examples are included in the resources folder.
I know it's a bit more cumbersome, but it allows the test cases to form a living documentation that can quickly be consulted to see what code gets generated for what API spec

ttobollik: Good point. We have added an example directory and added the tests to the general generator tests.

ttobollik:
We added an CLI flag to the `--http-controller-opts` to opt in for authentication. @cjbooms please review the branch when it is convenient for you.

cjbooms:
Nice work, this is a really high quality contribution.

cjbooms:
👍

cjbooms:
The only thing missing is to regenerate the usage instructions and update the README.
```
./gradlew printCodeGenUsage
```

PR #189, opened 2023-04-09
Support json merge patch

### LINKED ISSUE #184: Add support for JSON Merge Patch
JSON Merge Patch ([RFC 7386](https://www.rfc-editor.org/rfc/rfc7386)) is a standard request body format for PATCH endpoints. It defines how to partially update an existing entity, by distinguishing values between explicit "null"s and unspecified properties. For example, the body `{}` would describe a no-op, where as the body `{'name'=null}` would describe the operation to "delete" the `name` property of the entity.

To my knowledge, https://github.com/OpenAPITools/jackson-databind-nullable is a good library to support this kind of behavior in Java/Kotlin. It would be nice to have this supported by Fabrikt too.

The "proper" solution would be to check for media type `application/merge-patch+json` and generate the model classes appropriately. However, this would couple model generation and controller generation, which, to my knowledge, is not something Fabrikt was made for so far. 

So, the "easy" solution could be to add a custom extension to type schemas, like `x-merge-patch: true` and generate the model classes for those types to be used by the JSON Merge Patch pattern.

atollk:
Fixes #184 

I added support for an extension called `x-json-merge-patch`. Schemas annotated with that extension wrap all their fields in a `JsonNullable`. The effect applies to all nested objects as well but not to `$ref`ed types. This is the usual behavior as described in RFC 7386 but also allows for custom behavior, should the need exist.

I'm not sure why or how one would combine this feature with discriminator-based polymorphism, so I didn't write any tests for it, but in theory, I added support for that as well.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
@@ -4,8 +4,15 @@ import com.cjbooms.fabrikt.generators.JavaxValidationAnnotations.fieldValid
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata
 import com.cjbooms.fabrikt.model.KotlinTypeInfo
 import com.cjbooms.fabrikt.model.PropertyInfo
-import com.squareup.kotlinpoet.*
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.KModifier
+import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.asTypeName
 
 data class ClassSettings(
     val polymorphyType: PolymorphyType,
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
@@ -4,38 +4,43 @@ import com.cjbooms.fabrikt.generators.JavaxValidationAnnotations.fieldValid
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata
 import com.cjbooms.fabrikt.model.KotlinTypeInfo
 import com.cjbooms.fabrikt.model.PropertyInfo
-import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.FunSpec
-import com.squareup.kotlinpoet.KModifier
-import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.*
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
-import com.squareup.kotlinpoet.PropertySpec
-import com.squareup.kotlinpoet.TypeName
-import com.squareup.kotlinpoet.TypeSpec
-import com.squareup.kotlinpoet.asTypeName
 
-enum class ClassType {
-    VANILLA_MODEL,
-    SUPER_MODEL,
-    SUB_MODEL
+data class ClassSettings(
+    val polymorphyType: PolymorphyType,
+    val isMergePatchPattern: Boolean,
+) {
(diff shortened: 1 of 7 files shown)


cjbooms:
Would this be better achieved using the content type instead of an extension?
```
      requestBody:
        required: true
        content:
          application/merge-patch+json:
            schema:
              $ref: "#/components/schemas/Resource"
```

atollk:
@cjbooms In theory, yes, but in practice, I feel like you'd have to make too many assumptions. You'd have to run a discovery for all usages of the schema in controllers. But the real problem starts when you start using the same schema for both `merge-patch+json` and normal `json`. Then you'd have to generate multiple variants of the schema and all nested schemas as well.

cjbooms:
> @cjbooms In theory, yes, but in practice, I feel like you'd have to make too many assumptions. You'd have to run a discovery for all usages of the schema in controllers. But the real problem starts when you start using the same schema for both `merge-patch+json` and normal `json`. Then you'd have to generate multiple variants of the schema and all nested schemas as well.

I suppose in a future extension to this feature we could pre scan all schema objects referenced from that content-type and add the `x-json-merge-patch` key pre generation

src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
```diff
@@ -4,38 +4,43 @@ import com.cjbooms.fabrikt.generators.JavaxValidationAnnotations.fieldValid
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata
 import com.cjbooms.fabrikt.model.KotlinTypeInfo
 import com.cjbooms.fabrikt.model.PropertyInfo
-import com.squareup.kotlinpoet.ClassName
-import com.squareup.kotlinpoet.FunSpec
-import com.squareup.kotlinpoet.KModifier
-import com.squareup.kotlinpoet.ParameterSpec
+import com.squareup.kotlinpoet.*
```
cjbooms, line 7: revert wildcard import change

src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
```diff
@@ -2,30 +2,24 @@ package com.cjbooms.fabrikt.generators.model
...
 import com.cjbooms.fabrikt.generators.TypeFactory.createMapOfMapsStringToStringAny
 import com.cjbooms.fabrikt.generators.TypeFactory.createMapOfStringToType
 import com.cjbooms.fabrikt.generators.TypeFactory.createMutableMapOfMapsStringToStringType
 import com.cjbooms.fabrikt.generators.TypeFactory.createMutableMapOfStringToType
-import com.cjbooms.fabrikt.generators.JavaxValidationAnnotations
 import com.cjbooms.fabrikt.generators.ValidationAnnotations
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata.JSON_VALUE
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata.basePolymorphicType
 import com.cjbooms.fabrikt.generators.model.JacksonMetadata.polymorphicSubTypes
+import com.cjbooms.fabrikt.model.*
```
cjbooms, line 19: revert wild card import change

cjbooms:
I'm open to merging this feature, thanks for the contribution. Two things:
1. Should we add a new `http-model-opts` CLI parameter to allow users to opt-in?
2. If not doing `1.` which is self documentating, please think of a way to document this feature in the README

src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
atollk: done. Sorry, they were introduced by my IntelliJ

src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
atollk: done

atollk:
> I'm open to merging this feature, thanks for the contribution. Two things:
> 
>     1. Should we add a new `http-model-opts` CLI parameter to allow users to opt-in?
> 
>     2. If not doing `1.` which is self documentating, please think of a way to document this feature in the README

I don't think a CLI option is necessary here. This is not a breaking change and it already can be turned off conditionally by just not using the extension in OAS.

I added a bullet point to the README.md. I can see the potential for a more thorough feature documentation in general (I actually wanted to open an issue for this because I was struggling with the current polymorphism a lot). But, if that's a goal, it seems to be out of scope for this change.

PR #192, opened 2023-04-11
Added support for both `mapping` and `anyOf` for polymorphy.

### LINKED ISSUE #191: Allow `anyOf` with polymorphism
Currently, the way to use discriminator-based polymorphism with Fabrikt is:

```yaml
    polymorphicEnumDiscriminator:
      type: object
      discriminator:
        propertyName: some_enum
        mapping:
          obj_one_only: '#/components/schemas/ConcreteImplOneLegacy'
          obj_two_first: '#/components/schemas/ConcreteImplTwoLegacy'
          obj_two_second: '#/components/schemas/ConcreteImplTwoLegacy'
          obj_three: '#/components/schemas/ConcreteImplThreeLegacy'
      properties:
        some_enum:
          $ref: '#/components/schemas/EnumDiscriminator'
```

However, from all sources I could find, this isn't actually correct.
https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

You actually need to specify the possible "subschemas" with `anyOf` or `oneOf`. That is:

```yaml
    polymorphicEnumDiscriminator:
      type: object
      discriminator:
        propertyName: some_enum
        mapping:
          obj_one_only: '#/components/schemas/ConcreteImplOne'
          obj_two_first: '#/components/schemas/ConcreteImplTwo'
          obj_two_second: '#/components/schemas/ConcreteImplTwo'
          obj_three: '#/components/schemas/ConcreteImplThree'
      anyOf:
        - $ref: '#/components/schemas/ConcreteImplOne'
        - $ref: '#/components/schemas/ConcreteImplTwo'
        - $ref: '#/components/schemas/ConcreteImplTwo'
        - $ref: '#/components/schemas/ConcreteImplThree'
      properties:
        some_enum:
          $ref: '#/components/schemas/EnumDiscriminator'
```

This is also highlighted by the fact that many other OAS generators do not properly process the first variant. For example, https://editor.swagger.io/ renders it as an object with only a single property:

image

At the moment, the second variant causes a StackOverflowError in Fabrikt.

atollk:
Fixes #191 

The solution is to remove all `anyOf` from a model that has a `discriminator` and `mapping` defined. Not the prettiest solution but I couldn't see any other without re-thinking the entire approach.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/model/SourceApi.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/model/SourceApi.kt
@@ -49,26 +49,34 @@ data class SourceApi(
         val schemaErrors = api.schemas.entries.fold(emptyList<ValidationError>()) { errors, entry ->
             val name = entry.key
             val schema = entry.value
-            if (schema.type == OasType.Object.type && (
-                schema.oneOfSchemas?.isNotEmpty() == true ||
-                    schema.allOfSchemas?.isNotEmpty() == true ||
-                    schema.anyOfSchemas?.isNotEmpty() == true
-                )
-            )
-                errors + listOf(
+
+            val newError =
+                if (schema.type == OasType.Object.type) {
+                val hasOneOf = schema.oneOfSchemas?.isNotEmpty() == true
+                val hasAnyOf = schema.anyOfSchemas?.isNotEmpty() == true
+                val hasAllOf = schema.allOfSchemas?.isNotEmpty() == true
+                val hasMapping = schema.discriminator.hasMappings()
+                if (hasMapping && !hasOneOf && !hasAllOf) {
+                    // Polymorphism currently is only supported via the 'mapping' property.
+                    // All 'anyOf' schemas have to be removed here to avoid stack overflows later.
+                    schema.anyOfSchemas.clear()
+                    null
+                } else if (hasOneOf || hasAnyOf || hasAllOf) {
                     ValidationError(
                         "'$name' schema contains an invalid combination of properties and `oneOf | anyOf | allOf`. " +
-                            "Do not use properties and a combiner at the same level."
+                                "Do not use properties and a combiner at the same level."
                     )
-                )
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/cjbooms/fabrikt/model/SourceApi.kt
```diff
@@ -49,26 +49,34 @@ data class SourceApi(
...
+            val newError =
+                if (schema.type == OasType.Object.type) {
+                val hasOneOf = schema.oneOfSchemas?.isNotEmpty() == true
+                val hasAnyOf = schema.anyOfSchemas?.isNotEmpty() == true
+                val hasAllOf = schema.allOfSchemas?.isNotEmpty() == true
+                val hasMapping = schema.discriminator.hasMappings()
+                if (hasMapping && !hasOneOf && !hasAllOf) {
+                    // Polymorphism currently is only supported via the 'mapping' property.
+                    // All 'anyOf' schemas have to be removed here to avoid stack overflows later.
+                    schema.anyOfSchemas.clear()
```
cjbooms, line 62: Does this issue only happen with anyOf, or does it happen with allOf or oneOf also?

atollk: `allOf` doesn't support polymorphism at all, so that's no concern.
`oneOf` does, in theory, but from what I can tell, Fabrikt does not in practice. If Fabrikt were to support `oneOf` with discriminators at one point, the same issue might arise again.

atollk: I probably need to adapt this, now that #125 has been merged.

atollk: I updated the branch. Any more changes you want me done, @cjbooms ?

atollk:
Since there hasn't been any progress here and I'm not part of the Fabrikt-related project anymore, I'm going to unsubscribe from this PR.

cjbooms:
Thank you for your contribution @atollk and I apologise it was not accepted.

PR #193, opened 2023-04-11
General dependency update (#181)

cjbooms:
Comment below PR merged to this branch.  

FAILING, original author was not correct about it passing 

> I am interested in taking on issue #89, to support `kotlinx.serialization`.
>
> In preparation, I checked out `master` and observed that dependencies were generally somewhat out-of-date.  I've updated them here before adding support for the latest `kotlinx.serialization` version.
>
> A couple of minor changes arose to account for the new dependencies:
>
> 1. Dokka now prefers to expose a task-per-output-format, than have a single output format selected.  I imagine this is to support output of multiple formats.  The way the task is acquired for the `kotlinDocJar` task then had to change.
> 2. Kotlin Poet now supports multple 'simpleName's; I'm not sure why to be honest but fixing our usage was a simple case of removing the single argument label.
>
> Unit tests passing locally ✅

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-java@v1
       with:
-        java-version: '11'
+        java-version: '17'
 
     # Runs a single command using the runners shell
     - name: Build Project
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-java@v1
         with:
-          java-version: '11'
+          java-version: '17'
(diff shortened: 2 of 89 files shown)


zettelmj:
[test-fixes.patch.zip](https://github.com/cjbooms/fabrikt/files/12643117/test-fixes.patch.zip)

Looks like code generation has changed - loads of publics are getting added and keywords have to be escaped now...

cjbooms:
> [test-fixes.patch.zip](https://github.com/cjbooms/fabrikt/files/12643117/test-fixes.patch.zip)
> 
> Looks like code generation has changed - loads of publics are getting added and keywords have to be escaped now...

Yep, still probably worth updating all the tests examples and getting fabrikt up to latest lib and language versions. I just never get around to it

zettelmj:
The attached patch fixes the build. I can take over the PR and rebase it onto master

Issue #153, opened 2022-11-14
Add support for nested polymorphic models


averabaq:
It should be possible to generated nested polymorphic models. A schema could be both, SuperType and SubType, however only one is actual resolving

Example: 

`Given the API definition below`
```
  schemas:
    RootDiscriminatorType:
      type: string
      enum:
        - firstLevelChild

    FirstLevelDiscriminatorType:
      type: string
      enum:
        - secondLevelChild1
        - secondLevelChild2

    RootType:
      type: object
      required:
        - rootDiscriminator
      properties:
        rootDiscriminator:
          $ref: "#/components/schemas/RootDiscriminatorType"
        rootField1:
          type: string
        rootField2:
          type: boolean
      discriminator:
        propertyName: rootDiscriminator
        mapping:
          firstLevelChild: '#/components/schemas/FirstLevelChild'

    FirstLevelChild:
      allOf:
        - $ref: '#/components/schemas/RootType'
        - type: object
          required:
            - firstLevelDiscriminator
          properties:
            firstLevelDiscriminator:
              $ref: '#/components/schemas/FirstLevelDiscriminatorType'
            firstLevelField1:
              type: string
            firstLevelField2:
              type: integer
          discriminator:
            propertyName: firstLevelDiscriminator
            mapping:
              secondLevelChild1: '#/components/schemas/SecondLevelChild1'

    SecondLevelChild1:
      allOf:
        - $ref: '#/components/schemas/FirstLevelChild'
        - type: object
          required:
            - metadata
          properties:
            metadata:
              $ref: '#/components/schemas/SecondLevelMetadata'
```

`I would expect the model hierarchy to be`
```
RootType->FirstLevelChild -> SecondLevelChild
```

`Actual generation`

```
RootType -> FirstLevelChild
SecondLevelChild
```

averabaq:
Fixed by #195

PR #195, opened 2023-04-12
Add support for generating nested polymorphic models (#153)

averabaq:
Add support for generating nested polymorphic models 
Closes (#153)

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
@@ -27,6 +27,7 @@ data class ClassSettings(
 
 object PropertyUtils {
     fun PropertyInfo.addToClass(
+        modelName: String,
         type: TypeName,
         parameterizedType: TypeName,
         classBuilder: TypeSpec.Builder,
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
```diff
@@ -68,16 +69,17 @@ object PropertyUtils {
...
+                        if (discriminators.size == 1) {
+                            when (val discriminator = discriminators.first()) {
+                                is PropertyInfo.DiscriminatorKey.EnumKey ->
+                                    property.initializer("%T.%L", type, discriminator.enumKey)
+
+                                is PropertyInfo.DiscriminatorKey.StringKey ->
+                                    property.initializer("%S", discriminator.stringValue)
                             }
+                        } else {
+                            property.addAnnotation(JacksonMetadata.jacksonParameterAnnotation(oasKey))
```
cjbooms, line 82: So this else is for 0 and > 1? Seems a little strange...

averabaq: Yeah, that got me mad, but couldn't find a better way to handle this

That's essentially for covering this case 👉 https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/enumPolymorphicDiscriminator/models/Models.kt#L29

1) If there's no discriminator (size 0) [the property will go to the constructor](https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/enumPolymorphicDiscriminator/models/Models.kt#L13) with the `@param` annotation 
2) If there's only one discriminator (size 1) [the property will be a field with the discriminator value](https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/enumPolymorphicDiscriminator/models/Models.kt#L25) but without the `@param` annotation 
3) But if there are [many discriminators ](https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/enumPolymorphicDiscriminator/api.yaml#L19-L20) for the model class, then the property will have to go to the constructor too without any discriminator value as we can't chose, same as 1)

cjbooms:
I think the best thing to do is for me to merge this next week and cut a pre-release. Then test that pre-release against a couple of more complex APIs at Zalando to make sure it doesn't break anything. If all good, we can promote to maven central

PR #201, opened 2023-04-22
Improve generation of inlined objects

### LINKED ISSUE #188: Support generation of array models
Currently, defining schemas as `type: array` does not correctly generate model types.

For example:
```yml
    Foo:
      type: array
      items:
        type: string
        enum:
          - X
          - Y
```

I would expect this to generate an enum type named something like `FooItem` and represent references to `Foo` as `List<FooItem>`. 

What actually happens is that references are represented as `List<Foo>` but no type `Foo` is generated.

cjbooms:
Adds support  for top level array objects containing inlined items definitions.
Closes #188

### DIFF
--- a/src/test/resources/examples/enumExamples/models/Models.kt
+++ b/src/test/resources/examples/enumExamples/models/Models.kt
@@ -2,9 +2,16 @@ package examples.enumExamples.models
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.annotation.JsonValue
+import javax.validation.Valid
 import kotlin.String
 import kotlin.collections.List
 import kotlin.collections.Map
+
+data class BarBar(
+    @param:JsonProperty("bar_prop")
+    @get:JsonProperty("bar_prop")
+    val barProp: String? = null
+)
 
 enum class ContentType(
     @JsonValue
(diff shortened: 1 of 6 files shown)


src/test/resources/examples/enumExamples/models/Models.kt
```diff
@@ -2,10 +2,17 @@ package examples.enumExamples.models
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import com.fasterxml.jackson.annotation.JsonValue
+import javax.validation.Valid
 import kotlin.String
 import kotlin.collections.List
 import kotlin.collections.Map
 
+data class BarBar(
```
cjbooms, line 10: I don't see an easy way to not generate these. I think we may want them sometimes

src/test/resources/examples/enumExamples/models/Models.kt
```diff
@@ -128,3 +135,49 @@ enum class ExtensibleEnumObject(
...
+    @param:JsonProperty("prop_one")
+    @get:JsonProperty("prop_one")
+    val propOne: List<FooBarsFoo>? = null,
+    @param:JsonProperty("prop_two")
+    @get:JsonProperty("prop_two")
+    @get:Valid
+    val propTwo: List<FooBarsBar>? = null
+)
+
+data class FooBarsBar(
```
atollk, line 149: **suggestion:** Would it make more sense to add some version of "Item" to the generated name? I feel like just concatenating schema name and property name has a higher potential for conflicts.

cjbooms: It would be nice, but would be a larger refactor to make it play nice with Controller references

PR #205, opened 2023-05-02
Add support for requestBodies section. Fixes #20.

### LINKED ISSUE #20: Support generation of models from requestBodies section
Currently we only generate models from the schemas section

Consider expanding generation to support generating models from definitions within requestBodies or parameters sections too

atollk:
I added generation for models from the requestBodies section. 

I also changed a few dependencies from testImplementation to testCompileOnly to make it clear that they're not used for actually running the tests and applied a minor change which causes the media types of annotations on controller functions to be unique (i.e. you won't have produces = ["application/json", "application/json", "application/json"] anymore)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,19 +53,19 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter-engine:5.1.0")
     testImplementation("org.junit.jupiter:junit-jupiter-params:5.1.0")
     testImplementation("org.assertj:assertj-core:3.14.0")
-
-    // Below dependencies are solely present so code examples in the test resources dir compile
-    testImplementation("javax.validation:validation-api:2.0.1.Final")
-    testImplementation("jakarta.validation:jakarta.validation-api:3.0.2")
-    testImplementation("org.springframework:spring-webmvc:5.1.9.RELEASE")
-    testImplementation("org.springframework.security:spring-security-web:5.4.0")
-    testImplementation("io.micronaut:micronaut-core:3.8.4")
-    testImplementation("io.micronaut:micronaut-http:3.8.4")
-    testCompileOnly("io.micronaut.security:micronaut-security:3.8.3")
-    testImplementation("com.squareup.okhttp3:okhttp:4.9.1")
     testImplementation("com.pinterest.ktlint:ktlint-core:0.41.0")
     testImplementation("com.pinterest:ktlint:0.41.0")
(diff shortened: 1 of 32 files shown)


atollk:
@cjbooms I think that's not supported by Fabrikt at the moment anyway, am I wrong?
For example, this spec:

```yaml
openapi: 3.0.0
paths:
  /foo:
    post:
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/A'
          application/json2:
            schema:
              $ref: '#/components/schemas/B'
```

generates this Spring controller:

```kotlin
interface FooController {
  /**
   *
   *
   * @param a 
   * @param b 
   */
  @RequestMapping(
    value = ["/foo"],
    produces = ["application/json"],
    method = [RequestMethod.POST],
    consumes = ["application/json", "application/json2"]
  )
  fun post(@RequestBody @Valid a: A, @RequestBody @Valid b: B): ResponseEntity<String>
}
```

which, to my knowledge, is invalid code.

cjbooms:
But for users of model generation only, this change has the potential to break them.

atollk:
Ah, I see what you mean. So I could either:
- change the name of the generated classes, if there are multiple; or
- only generate the class for the first media type; or
- add a CLI flag and only generate request bodies conditionally if set.

Which would you consider best?

pschichtel:
This is probably not a good place to discuss this, but just my 2 cents:

When I faced the issue of having one operation with multiple different response types using the openapi-generator, I attempted to customize the controller template to generate a dedicated controller method for each combination of request body type and response body type. It eventually failed with kotlin (or actually java) not allowing `/` in function names and mustache templates being too limited.

At least for spring I still think it's the right way to encode this, since spring will do the correct matching based on the Accept and Content-Type headers and each method can easily be typed using the appropriate models.

cjbooms:
I think the first option would be best, either changing the name when there are multiple or always taking some input from content key. 

We can look at adding support to the controller generation later.

One other thing that we'd have to decide is if there is any point in generating a model class for request body if its schema is simply a $ref to another schema

atollk:
> I think the first option would be best, either changing the name when there are multiple or always taking some input from content key.

Done. I just added index suffixes for now. 

> One other thing that we'd have to decide is if there is any point in generating a model class for request body if its schema is simply a $ref to another schema

I don't think I have a strong opinion on this one but intuitively I'd say yes.

cjbooms:
Merging to branch so I can make edits

PR #208, opened 2023-05-08
Fix nullability and default value issues for JSON merge patch

atollk:
Introduced in #189 , the "JSON Merge Patch" pattern implemented via `JsonNullable` still has some rough edges like bugs with nullability of the inner type parameters as well as with default values of properties. Both should be fixed with this PR.


larsduelfer:
Would be awesome if this can be integrated soon since we are relying on this fix! Thanks in advance!

cjbooms:
👍

cjbooms:
released now, and should sync to maven central in the next 30 minutes

larsduelfer:
Perfect, thanks @cjbooms

PR #215, opened 2023-05-18
Issue 20

cjbooms:
From @atollk, with the minor change of the suffix to make it more intuitive.

> I added generation for models from the requestBodies section.
>
> I also changed a few dependencies from testImplementation to testCompileOnly to make it clear that they're not used for actually running the tests and applied a minor change which causes the media types of annotations on controller functions to be unique (i.e. you won't have produces = ["application/json", "application/json", "application/json"] anymore)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -53,19 +53,19 @@ dependencies {
     testImplementation("org.junit.jupiter:junit-jupiter-engine:5.1.0")
     testImplementation("org.junit.jupiter:junit-jupiter-params:5.1.0")
     testImplementation("org.assertj:assertj-core:3.14.0")
-
-    // Below dependencies are solely present so code examples in the test resources dir compile
-    testImplementation("javax.validation:validation-api:2.0.1.Final")
-    testImplementation("jakarta.validation:jakarta.validation-api:3.0.2")
-    testImplementation("org.springframework:spring-webmvc:5.1.9.RELEASE")
-    testImplementation("org.springframework.security:spring-security-web:5.4.0")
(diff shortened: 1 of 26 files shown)


cjbooms:
@atollk We'll have to work on fixing the broken controller references if we want to merge this.

cjbooms:
There is still plenty of work to do here. In particular, all of these controller functions should probably be switched to returning  the response and request body variants
https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/githubApi/api.yaml#L828-L858

cjbooms:
This feature is so brittle with how it integrates with Controller interfaces and Client gen, I'm considering not implmenting it.

That, or perhaps remove client generation entirely from this code base

atollk:
Could you help me understand what the purpose of your commits on this branch is? Was the code generation of my change alone errorneous or is it purely cosmetic?

If it's the latter, I agree with the underlying motivation that appending indices to the model class names in some cases isn't a great solution but I feel like it would be sufficient to resolve this issue and officially have "request bodies" supported. Also because from what I can tell, your added changes now make this PR not backwards-compatible due to different naming of response schemas.

For a general solution, I feel like model class mappings deserve a refactoring anyway. But that should be handled in a different issue imo.

cjbooms:
If we start generating models from requestBodies or responses section, these models should be used in the Controller interfaces and client implementations.

But the logic around this is super-brittle and relies on the same name being chosen at generation time in separate code paths. I attempted to fix this, but ran out of time

PR #217, opened 2023-06-13
Move discriminated field to constructor.

cjbooms:
This is necessary even when there is only one value to have it used as part of the equals function


cjbooms:
👍

ePaul:
:+1:

PR #220, opened 2023-06-14
Add missing JsonProperty parameter annotation.

### LINKED ISSUE #219: Polymorphism Regression in 9.0
The following spec snippet results in two data classes `ArticleSourceReference` and `ArticleIdReference` where the property for the discriminator is now a constructor argument which leads to a data bind exception at runtime.

```yaml
    ReferenceDiscriminatorType:
      type: string
      enum:
        - Id
        - Source

    ArticleReference:
      discriminator:
        propertyName: type
        mapping:
          Source: '#/components/schemas/ArticleSourceReference'
          Id: '#/components/schemas/ArticleIdReference'
      required:
        - type
      properties:
        type:
          $ref: '#/components/schemas/ReferenceDiscriminatorType'

    ArticleSourceReference:
      allOf:
        - $ref: '#/components/schemas/ArticleReference'
        - type: object
          required:
            - provenance
            - sourceId
          properties:
            provenance:
              type: string
            sourceId:
              type: string

    ArticleIdReference:
      allOf:
        - $ref: '#/components/schemas/ArticleReference'
        - type: object
          required:
            - id
          properties:
            id:
              type: string
```

### Expected Behaviour

Fabrikt 9.0 should generate the same as 8.x, where the discriminator property `type` is not part of the constructor.
```kotlin
data class ArticleIdReference(
    @param:JsonProperty("id")
    @get:JsonProperty("id")
    @get:NotNull
    val id: String
) : ArticleReference() {
    @get:JsonProperty("type")
    @get:NotNull
    override val type: ReferenceDiscriminatorType = ReferenceDiscriminatorType.ID
}
```

### Actual Behaviour
Fabrikt 9.0 generates data classes where the discriminator property `type` is a constructor argument.
```kotlin
data class ArticleIdReference(
    @param:JsonProperty("id")
    @get:JsonProperty("id")
    @get:NotNull
    val id: String,
    @get:JsonProperty("type")
    @get:NotNull
    override val type: ReferenceDiscriminatorType = ReferenceDiscriminatorType.ID
) : ArticleReference()
```

This compiles but when used with Quarkus and Jackson I get the following databind exception:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `xyz.rest.models.ArticleSourceReference`: Argument #2 of constructor [constructor for `xyz.rest.models.ArticleSourceReference` (3 args), annotations: [null] has no property name annotation; must have name when multiple-parameter constructor annotated as Creator
```

cjbooms:
Since moving the discriminator property from the class into the constructor, we need to have the parameter annotation also. Closes #219

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
@@ -108,6 +108,7 @@ object PropertyUtils {
                 this as PropertyInfo.Field
                 if (classSettings.polymorphyType == ClassSettings.PolymorphyType.SUB) {
                     property.initializer(name)
+                    property.addAnnotation(JacksonMetadata.jacksonParameterAnnotation(oasKey))
                     val constructorParameter: ParameterSpec.Builder = ParameterSpec.builder(name, wrappedType)
                     val discriminators = maybeDiscriminator.getDiscriminatorMappings(schemaName)
                     when (val discriminator = discriminators.first()) {
--- a/src/test/resources/examples/companionObject/models/Models.kt
+++ b/src/test/resources/examples/companionObject/models/Models.kt
@@ -31,6 +31,7 @@ data class Cat(
     val age: Int? = null,
     @get:JsonProperty("petType")
     @get:NotNull
+    @param:JsonProperty("petType")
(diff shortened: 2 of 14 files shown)


cjbooms:
👍

PR #221, opened 2023-07-05
Remove redundant checks for co-occurrence of properties and combiners

rajki:
#### Context
Was browsing through the codebase and found these two redundant conditions. Unless I'm missing something, these conditions are already subsumed in the `if` block above, correct? We could remove them.


cjbooms:
Thanks. Strange that my IntelliJ didn't spot that

PR #222, opened 2023-07-10
Feature/client open feign

dpnolte:
Hi there,

We really like FabriKt ❤️!  But we were missing our preferred client, OpenFeign. 
This PR therefore introduces the support OpenFeign as a client. To be more specific, this PR enables to generate interfaces for OpenFeign targets like:
```
@Suppress("unused")
interface ExamplePath1Client {
    /**
     * GET example path 1
     *
     * @param explodeListQueryParam
     * @param queryParam2
     */
    @RequestLine("GET /example-path-1?explode_list_query_param={explodeListQueryParam}&query_param2={queryParam2}")
    @Headers("Accept: application/vnd.custom.media+json")
    fun getExamplePath1(
        @Param("explodeListQueryParam") explodeListQueryParam: List<String>? = null,
        @Param("queryParam2") queryParam2: Int? = null,
        @HeaderMap additionalHeaders: Map<String, String> = emptyMap()
    ): QueryResult
}
```

It would then be up to the consumer of this library to build the OpenFeign client and set this interface as a target.  

To determine which client needs to be generated, I've followed the option structure as with controllers:
image

Since the generated interface does not include the client itself, I didn't add the support yet for `RESILIENCE4J ` with OpenFeign. I did add a new option for OpenFeign to have the interface methods generated with suspend modifier so that the CoroutineFeign could be used ([reference](https://github.com/OpenFeign/feign/tree/master/example-github-with-coroutine)).

This PR also addresses an issue that we found. If you had a different reponse content types for success and error repsonses, Fabrikt would set the return type of the client method to JsonNode. I understand that this necessary when you have multiple reponse content type if the status is in the successful range. However, when the response is not in the successful range, both clients will throw an error (i.e., `ApiException` and `FeignException`). I've therefore adjusted it to only set the response type to `JsonNode` if multiple content types are found for successful responses (see commit https://github.com/cjbooms/fabrikt/pull/222/commits/4b65287466b6c2ce0449a80e34aac0287effc4d3).
 I've also added a test case for this `differentErrorResponse`.


src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenArgs.kt
```diff
@@ -106,6 +106,13 @@ class CodeGenArgs {
     )
     var clientOptions: Set<ClientCodeGenOptionType> = emptySet()
 
+    @Parameter(
+        names = ["--http-client-target"],
+        description = "Optionally select the target client that you want to be generated. Defaults to Spring Controllers",
```
cjbooms, line 111:
```suggestion
        description = "Optionally select the target client that you want to be generated. Defaults to OK_HTTP",
```

README.md
```diff
@@ -186,48 +186,53 @@ data class Responses(
...
-|   `--type-overrides`         | Specify non-default kotlin types for certain OAS types. For example, generate `Instant` instead of `OffsetDateTime` |
-|                              | CHOOSE ANY OF: |
-|                              |   `DATETIME_AS_INSTANT` - Use `Instant` as the datetime type. Defaults to `OffsetDateTime` |
-|                              |   `DATETIME_AS_LOCALDATETIME` - Use `LocalDateTime` as the datetime type. Defaults to `OffsetDateTime` |
-|   `--validation-library`     | Specify which validation library to use for annotations in generated model classes. Default: JAVAX_VALIDATION |
-|                              | CHOOSE ONE OF: |
-|                              |   `JAVAX_VALIDATION` - Use `javax.validation` annotations in generated model classes (default) |
-|                              |   `JAKARTA_VALIDATION` - Use `jakarta.validation` annotations in generated model classes |
+| Parameter                | Description                                                                                                                                                                |
```
cjbooms, line 189: Was this updated table generated via: `./gradlew printCodeGenUsage` ?

src/test/resources/examples/differentErrorResponseBody/api.yaml
```diff
@@ -0,0 +1,32 @@
+openapi: 3.0.0
```
cjbooms, line 1: I think it would be cleaner if we just incorporated this additional test case into the existing example:
`okHttpClientMultiMediaType`

cjbooms:
This is a really high quality contribution. Kudos to all involved. 💯

README.md
dpnolte: aah wow cool! I'll replace the table with the generated version

dpnolte:
Thanks @cjbooms!! It is also a joy to with such nicely structured library 🚀 . I also think that using KotlinPoet is really smart to reduce the risk of having invalid Kotlin code generated.

I've updated the PR as per your comments.

cjbooms:
👍

cjbooms:
LGTM - I can cut a release today and let you play around with it locally for a bit. If there are any issues we can cut new hot-fix releases until you are happy

dpnolte:
Awesome, thanks!  So far it is looking good from our side 🚀

PR #226, opened 2023-07-24
Support body on delete requests

### LINKED ISSUE #225: DELETE doesn't propagate the body
The HTTP DELETE method supports a body, as defined in https://www.rfc-editor.org/rfc/rfc7231#section-4.3.5:

>    A payload within a DELETE request message has no defined semantics;
   sending a payload body on a DELETE request might cause some existing
   implementations to reject the request.

But the code generator simply ignores it. It should be easy to fix by changing this line of code: https://github.com/cjbooms/fabrikt/blob/master/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt#L212

slinkydeveloper:
Fix #225

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt
@@ -209,7 +209,7 @@ data class SimpleClientOperationStatement(
             "PATCH" -> this.addRequestSerializerStatement("patch")
             "HEAD" -> this.add("\n.head()")
             "GET" -> this.add("\n.get()")
-            "DELETE" -> this.add("\n.delete()")
+            "DELETE" -> this.addRequestSerializerStatement("delete")
             else -> throw NotImplementedError("API operation $op is not supported")
         }
         return this.add("\n.build()\n")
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt
```diff
@@ -230,6 +230,6 @@ data class SimpleClientOperationStatement(
                 requestBody.getPrimaryContentMediaType()?.key,
                 "toMediaType".toClassName("okhttp3.MediaType.Companion")
             )
-        } ?: this.add("\n.%N(ByteArray(0).%T())", verb, toRequestBody)
+        } ?: if (verb == "delete") this.add("\n.delete()") else this.add("\n.%N(ByteArray(0).%T())", verb, toRequestBody)
```
aderington, line 233: This change seems out of place.  I suggest checking to see if a `delete` verb has any `Body` parameters in `addRequestStatement`.

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt
```diff
@@ -209,7 +209,7 @@ data class SimpleClientOperationStatement(
             "PATCH" -> this.addRequestSerializerStatement("patch")
             "HEAD" -> this.add("\n.head()")
             "GET" -> this.add("\n.get()")
-            "DELETE" -> this.add("\n.delete()")
+            "DELETE" -> this.addRequestSerializerStatement("delete")
```
aderington, line 212: I haven't cloned the repo yet, but on first light, I think something like this would be a more cohesive solution.

```
"DELETE" -> {
  val hasBodyParameters = parameters.any { it is BodyParameter }
  if (hasBodyParameters) {
      this.addRequestSerializerStatement("delete")
  } else {
       this.add("\n.delete()")
  }
}
```

cjbooms:
Closing this PR as contributor never responded

PR #230, opened 2023-08-28
Add "auto generated" comment on top of generated files

ReallyLiri:


cjbooms:
Hi @ReallyLiri this is not something I would like to add to fabrikt.
I suggest you instead add this comment to the generated classes via your build script, after the generate task has run. With gradle, this can be accomplished by adding a doLast section to your task

PR #231, opened 2023-08-31
Add support for type string with format binary

ascheja:
Right now trying to specify a schema like `{"type": "string", "format": "binary"}` falls back to `kotlin.String`.

The binary format is one of the string formats [defined by the openapi specification](https://swagger.io/docs/specification/data-models/data-types/#format), used mostly for describing endpoints accepting/returning [files](https://swagger.io/docs/specification/data-models/data-types/#file).

The last commit is actually optional. [`JsonOverlay` encodes `/` as `~1`](https://github.com/RepreZen/JsonOverlay/blob/master/json-overlay/src/main/java/com/reprezen/jsonoverlay/JsonOverlay.java#L203) before joining the path together with `/`, which leads to a strange to read parameter name like `application1octetStream`.

### DIFF
--- a/src/test/resources/examples/binary/api.yaml
+++ b/src/test/resources/examples/binary/api.yaml
@@ -1,0 +1,24 @@
+openapi: 3.0.1
+info:
+  description: Testing binary body and binary response
+  title: Test
+  version: '0.0'
+paths:
+  /binary-data:
+    post:
+      operationId: postBinaryData
+      requestBody:
+        required: true
+        content:
+          application/octet-stream:
+            schema:
+              type: string
+              format: binary
+      responses:
+        200:
+          description: Success
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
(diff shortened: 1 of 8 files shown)


src/test/resources/examples/binary/api.yaml
```diff
@@ -0,0 +1,24 @@
+openapi: 3.0.1
```
cjbooms, line 1: could we tack this example on to an existing test case rather than creating a new example?

ascheja: Sure, which one do you want it added to? From what I can see there are 5 files that currently generate controller code for both spring and micronaut:
- githubApi
- singleAllOf
- authentication
- parameterNameClash
- pathLevelParameters

Based on that I don't think it really fits anywhere?

cjbooms: ok, will merge as-is

PR #240, opened 2023-09-24
address a couple of compiler warnings and replace deprecated syntax

zettelmj:
address a couple of compiler warnings and replace deprecated syntax


cjbooms:
nice one. Thanks @zettelmj

PR #242, opened 2023-09-29
Quote enum item names when necessary

### LINKED ISSUE #241: Escape enum item name when it is not a valid identifier
Kotlin identifier should start with a letter or underscore.

Enum item name should be surrounded by backticks if it is not a valid Kotlin identifier.

```yaml
EnumObject:
  type: string
  enum:
    - 1one
    - 2two
```

should generate an enum with items like

```kotlin
EnumObject.`1ONE`
EnumObject.`2TWO`
```

detouched:
Add backticks to the enum item name in case it doesn't start with a letter or an underscore and hence is not a valid Kotlin identifier.

Fixes #241.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/OasDefault.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/OasDefault.kt
@@ -1,7 +1,7 @@
 package com.cjbooms.fabrikt.generators
 
 import com.cjbooms.fabrikt.model.KotlinTypeInfo
-import com.cjbooms.fabrikt.util.toUpperCase
+import com.cjbooms.fabrikt.util.NormalisedString.toEnumName
 import com.squareup.kotlinpoet.ClassName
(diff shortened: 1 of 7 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/OasDefault.kt
```diff
@@ -39,7 +39,7 @@ sealed class OasDefault {
 
     data class EnumValue(val type: TypeName, val enumValue: String) : OasDefault() {
         override fun getDefault(): CodeBlock =
-            CodeBlock.of("%T.${enumValue.toUpperCase()}", type)
+            CodeBlock.of("%T.${enumValue.toEnumName()}", type)
```
detouched, line 42: When referencing the enum item, the name should be formed the same way as when that item is generated – using `toEnumName()` method.

cjbooms:
Nice work. Thank you. Will cut a release for you this morning

PR #245, opened 2023-10-13
fix: missing types from other spec file

dpnolte:
Hi 👋 

Still enjoying your great library 💯  And we are adding it more and more to our projects.

We've encountered one issue of an Open API Spec contained in multiple files via external references.
Currently, not all types are generated if the model is from an external spec file.
More specifically, models from external spec files are filtered out by:
```
.filter { apiSchema -> externalReferences.value.contains(apiSchema.name) }
```

We've found that it can cause issues in at least the following cases:

**1. Parameter contains  external reference** 
With a spec like,
```
paths:
  /hello:
    get:
      operationId: helloWorld
      parameters:
        - name: language
          in: path
          required: true
          schema:
            $ref: './external-models.yaml#/components/schemas/Language'
```
the `Language` type wouldn't be created as it is currently filtered out.

***2. Schema in external file contains reference to another schema in the external file***
When an external spec file, contains another reference in the same file it will also be filtered out. That is, suppose that defined `Content` is referenced from `api.yaml` but defined in `external-models.yaml` like so:
```
    Content:
      type: object
      properties:
        contentType:
          type: string
          enum:
            - HTML
            - MD
      discriminator:
        propertyName:
          contentType
        mapping:
          HTML: '#/components/schemas/ContentHTML'
          MD: '#/components/schemas/ContentMD'
    ContentHTML:
      allOf:
        - $ref: '#/components/schemas/Content'
        - type: object
          properties:
            header:
              type: string
            body:
              type: string
    ContentMD:
      allOf:
        - $ref: '#/components/schemas/Content'
        - type: object
          properties:
            richText:
              type: string
```
Then `ContentHTML` and `ContentMD` will not be generated as it will be filtered out as well.

This PR therefore proposes to remove the filtering. The downside is that it generates more code, but at at least it would work in more scenarios. What do you think?

In any case, thanks again for this awesome library 👍

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
@@ -150,7 +150,6 @@ class JacksonModelGenerator(
             val api = OpenApi3Parser().parse(externalReferences.key)
             val schemas = api.schemas.entries.map { it.key to it.value }
                 .map { (key, schema) -> SchemaInfo(key, schema) }
-                .filter { apiSchema -> externalReferences.value.contains(apiSchema.name) }
             val externalModels = createModels(api, schemas)
             externalModels.forEach { additionalModel ->
                 if (models.none { it.name == additionalModel.name }) models.add(additionalModel)
--- a/src/test/kotlin/com/cjbooms/fabrikt/generators/OkHttpClientGeneratorTest.kt
+++ b/src/test/kotlin/com/cjbooms/fabrikt/generators/OkHttpClientGeneratorTest.kt
@@ -16,6 +16,7 @@ import com.cjbooms.fabrikt.model.SourceApi
 import com.cjbooms.fabrikt.util.Linter
 import com.cjbooms.fabrikt.util.ResourceHelper.readTextResource
 import com.squareup.kotlinpoet.FileSpec
+import java.nio.file.Paths
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.TestInstance
@@ -33,6 +34,7 @@ class OkHttpClientGeneratorTest {
         "okHttpClientPostWithoutRequestBody",
         "pathLevelParameters",
         "parameterNameClash",
+        "externalReferences"
     )
 
     @BeforeEach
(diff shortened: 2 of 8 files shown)


cjbooms:
Thanks for the detailed report and fix attempt. I'll try and take a look properly in next few days. 
I iterated on the external reference stuff quite extensively back in the day, it is very hard to get right in all circumstances, especially with unlined schema definitions come into the mix. 
IIRC correctly, we explicitly filtered them out for some reason, so we could generate them in isolation later. I can't remember full details, but will look at your code changes to see if I can remember any of the edge cases.

dpnolte:
Thanks for the quick reply! Let me know if I can help somehow to make this move forward. Perhaps it would be better to control filtering with a configuration flag?

dpnolte:
Based on your comments, I tried to make it it more robust by adding a `ModelNameRepository` that prevents name collisions for types. If a name collides, it will add a suffix to the name (e.g., "Extra"). I've added a case for this in the `externalReference` test case. I hope you will like it.

I've also added some of the complex inline type definitions to verify those work as well.  Up to this point, I couldn't find/reproduce the issue with respect to external files and inline schema definitions. If you recall, could you please let me know? I'd like to add a test case for that.

Cheers

cjbooms:
This is a really nice solution. Kudos.

It overlaps a little bit with the `--api-fragment` feature, but your code changes handles conflicting schema definitions.

We should make this feature opt-in for back-wards compatibility, and to address the following:
1. To allow builders to keep the amount of generated code to a minimum, especially when a smaller primary API references a singe schema in a much larger external API.
2. Today a primary API can references a single schema in an external API that contains one or more other bad schema definitions. With this change, the bad schemas will be imported and could result in generated code that does not compile. 

Perhaps a new one like
```
enum class ExternalReferencesResolutionMode(val description: String) {
    TARGETED("Generate models only for directly referenced schemas in external API files."),
    AGGRESSIVE("Generate models for every external schema when an external API is referenced.");

    override fun toString() = "`${super.toString()}` - $description"
}
```

dpnolte:
Thanks for the quick review and reply 💪 I really appreciate that. 

Your points makes totally sense to me as well.  I've updated the PR by adding the configuration flag as per your input.  I've also added separate test cases for ExternalReferenceResolutionMode `TARGETED` and `AGGRESSIVE`.

cjbooms:
👍

cjbooms:
Nice one. Given the code generation default hasn't changed. I'll release this now as a minor version

PR #247, opened 2023-10-23
Support schemas having properties with an "allOf with single $ref" type (#246)

### LINKED ISSUE #246: allOf with single $ref generates uncompilable code
I came across a specification that uses the "allOf with single $ref" pattern, most likely so they can put a description on properties with a referenced type. This works with Fabrikt as long as the property has the same name as the type, but fails if they are different. The following specification generates model code that does not compile.

```yaml
openapi: 3.0.3
info:
  title: "Dog API"
  version: "1"
paths: { }
components:
  schemas:

    Dog:
      type: object
      properties:
        owner:
          description: "The registered owner of the dog."
          allOf:
            - $ref: "#/components/schemas/Person"
        walker:
          description: "The person walking the dog."
          allOf:
            - $ref: "#/components/schemas/Person"

    Person:
      type: object
      required:
        - name
      properties:
        name:
          type: string
```

## Current Result
Fabrikt 11.0.0 generates the following two data classes:
```kotlin
// Dog.kt
public data class Dog(
  @param:JsonProperty("owner")
  @get:JsonProperty("owner")
  @get:Valid
  public val owner: Owner? = null,    // <- type Owner is not generated
  @param:JsonProperty("walker")
  @get:JsonProperty("walker")
  @get:Valid
  public val walker: Walker? = null,  // <- type Walker is not generated
)

// Person.kt
public data class Person(
  @param:JsonProperty("name")
  @get:JsonProperty("name")
  @get:NotNull
  public val name: String,
)
```
## Expected Result
I expect Fabrikt to generate compiling model code. 

One solution would be to treat properties with an allOf with a single $ref the same as properties that use the $ref directly, i.e. the generated Dog.kt would contain properties of type `Person`.
```kotlin
public data class Dog(
  @param:JsonProperty("owner")
  @get:JsonProperty("owner")
  @get:Valid
  public val owner: Person? = null,    // <- uses type Person instead of Owner
  @param:JsonProperty("walker")
  @get:JsonProperty("walker")
  @get:Valid
  public val walker: Person? = null,   // <- uses type Person instead of Walker
)
```
Another solution is to generate the data classes `Owner` and `Walker` which would look the same as the data class `Person`.

I prefer the first solution as it is simpler and represents the original intention of the specification.

acanda:
Previously, a property with an `allOf` type with a single `$ref` would generate a model property with a type derived from the name of the property instead of the type referenced by $ref. This fixes #246.

This commit implements the first solution mentioned in #246, i.e.

```yaml
    Dog:
      type: object
      properties:
        owner:
          description: "The registered owner of the dog."
          allOf:
            - $ref: "#/components/schemas/Person"
        walker:
          description: "The person walking the dog."
          allOf:
            - $ref: "#/components/schemas/Person"

    Person:
      type: object
      required:
        - name
      properties:
        name:
          type: string
```
will generate data classes where the properties `owner` and `walker` have type `Person` instead of `Owner` and `Walker`.
```kotlin
public data class Dog(
    @param:JsonProperty("owner")
    @get:JsonProperty("owner")
    @get:Valid
    public val owner: Person? = null,
    @param:JsonProperty("walker")
    @get:JsonProperty("walker")
    @get:Valid
    public val walker: Person? = null,
)

public data class Person(
    @param:JsonProperty("name")
    @get:JsonProperty("name")
    @get:NotNull
    public val name: String,
)
```

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
@@ -4,7 +4,6 @@ import com.cjbooms.fabrikt.cli.ModelCodeGenOptionType
 import com.cjbooms.fabrikt.generators.MutableSettings
 import com.cjbooms.fabrikt.model.OasType
 import com.cjbooms.fabrikt.model.PropertyInfo
-import com.cjbooms.fabrikt.util.NormalisedString.toMapValueClassName
 import com.cjbooms.fabrikt.util.NormalisedString.toModelClassName
 import com.reprezen.jsonoverlay.Overlay
 import com.reprezen.kaizen.oasparser.model3.Discriminator
@@ -189,6 +188,7 @@ object KaizenParserExtensions {
     fun Schema.safeName(): String =
         when {
             isOneOfPolymorphicTypes() -> this.oneOfSchemas.first().allOfSchemas.first().safeName()
+            isPropertyWithAllOfSingleType() -> this.allOfSchemas.first().safeName()
             name != null -> name
             else -> Overlay.of(this).pathFromRoot
                 .splitToSequence("/")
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
```diff
@@ -222,6 +222,26 @@ object KaizenParserExtensions {
...
 
+    private fun Schema.isPropertyWithAllOfSingleType() =
+        allOfSchemas?.size == 1 && isPropertySchema()
+
+    /**
+     * The `pathFromRoot` of a property schema ends with
+     * `/properties/<name of property>`, so we check if the
+     * penultimate segment is `properties`.
+     */
+    private fun Schema.isPropertySchema(): Boolean {
```
cjbooms, line 233: Could this be used instead?
`isInlinedObjectDefinition()`

cjbooms:
Nice work, thanks for the contribution. One minor nitpick, but happy to merge if you are

src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
acanda: `isInlinedObjectDefinition()` does not work for this case. `isObjectType()` returns false as the type of the schema `{"description":"The registered owner of the dog.","allOf":[{"$ref":"#/components/schemas/Person"}]}` is `null` and not `object`.

Would `isInlinedPropertySchema()` follow the naming convention or do you have something other in mind?

cjbooms: maybe just `isInlinedSchema()`

acanda: I renamed the function to `isInlinedPropertySchema()`.

PR #249, opened 2023-10-29
support kotlinx serialization

zettelmj:
support kotlinx serialization (see #89)


ulrikandersen:
Work continued in #327. Thanks for the draft 🙏

PR #252, opened 2023-11-24
Add module for end 2 end tests for verify the the generated client co…

zettelmj:
…mplies with the spec and is run able

### DIFF
--- a/end2end-tests/src/test/kotlin/com/cjbooms/fabrikt/clients/okio3/Okio3Test.kt
+++ b/end2end-tests/src/test/kotlin/com/cjbooms/fabrikt/clients/okio3/Okio3Test.kt
@@ -1,0 +1,269 @@
+package com.cjbooms.fabrikt.clients.okio3
+
+import com.example.client.*
+import com.example.models.Failure
+import com.example.models.FirstModel
+import com.example.models.QueryResult
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.github.tomakehurst.wiremock.WireMockServer
+import com.github.tomakehurst.wiremock.common.ConsoleNotifier
+import com.github.tomakehurst.wiremock.core.WireMockConfiguration.options
+import com.marcinziolo.kotlin.wiremock.*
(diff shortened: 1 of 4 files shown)


zettelmj:
@cjbooms I would propose to add acceptance/end2end tests like this.
As I am working on the jdk client this comes handy to make sure that both versions of the generated client code work the same way. For the new client I would like to run the same set of tests with both clients.

cjbooms:
> @cjbooms I would propose to add acceptance/end2end tests like this.
> As I am working on the jdk client this comes handy to make sure that both versions of the generated client code work the same way. For the new client I would like to run the same set of tests with both clients.

Yep, fine with me. Nice that it's a separate module

end2end-tests/src/test/kotlin/com/cjbooms/fabrikt/clients/okio3/Okio3Test.kt
```diff
@@ -0,0 +1,224 @@
...
+            urlPath like "/example-path-2/304"
+        } returns {
+            statusCode = 304
+        }
+
+        val result = assertThrows<ApiException> {
+            examplePath2Client.getExamplePath2PathParam("304", 10)
+        }
+
+        assertThat(result.message).isEqualTo("[304]: ")
```
zettelmj, line 220: @cjbooms This looks dodgy. There seems to be no handling for the 3xx cases.
Messages are either `successful -> 200..299`, `bad request -> 400..499` or `server error -> 500..599`. 
Everything 3xx falls through into the else case in `http-util.kt.hbs` line 46.

cjbooms: I don't think anyone at Zalando have used the API client for anything other than tests, so probably proper 3X handling never came up.

zettelmj:
First part of the coverage done

zettelmj:
#257

cjbooms:
👍

cjbooms (approved):

PR #254, opened 2023-11-29
Support 3xx http responses statuses in generated client code

### LINKED ISSUE #255: [client] Generic Exception is thrown when Server returns 3xx Response
In auto generated clients when the server returns a 3xx Http response the `else` part of the pattern match within the `HttpUtil.kt` code path is used. 
Instead a dedicated exception should be thrown.

zettelmj:
Support 3xx http responses statuses in generated client code

Fixes #255


cjbooms (approved):
LGTM

PR #259, opened 2024-01-22
Move additionalProperties map out of constructor

networkException:
Previously, fabrikt would generate a `properties` constructor argument when `additionalProperties` is `true` for an object schema. This would cause construction of such models when deserializing as jackson has no way of setting values in that map.

This pull request moves the map definition out of the constructor into a member, allowing the model to be constructed properly. Additional properties get inserted into and queried from the list using the already existing `@JsonAnySetter` and `@JsonAnyGetter` functions.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/PropertyUtils.kt
@@ -46,11 +46,8 @@ object PropertyUtils {
         val property = PropertySpec.builder(name, wrappedType)
 
         if (this is PropertyInfo.AdditionalProperties) {
-            property.initializer(name)
+            property.initializer("mutableMapOf()")
             property.addAnnotation(JacksonMetadata.ignore)
-            val constructorParameter: ParameterSpec.Builder = ParameterSpec.builder(name, wrappedType)
-            constructorParameter.defaultValue("mutableMapOf()")
-            constructorBuilder.addParameter(constructorParameter.build())
 
             val value =
                 if (typeInfo is KotlinTypeInfo.MapTypeAdditionalProperties) {
--- a/src/test/resources/examples/anyOfOneOfAllOf/models/Models.kt
+++ b/src/test/resources/examples/anyOfOneOfAllOf/models/Models.kt
@@ -81,9 +81,10 @@ public data class OneOfAdditionalProps(
     @param:JsonProperty("second_nested_any_of_prop")
     @get:JsonProperty("second_nested_any_of_prop")
     public val secondNestedAnyOfProp: String? = null,
-    @get:JsonIgnore
-    public val properties: MutableMap<String, Any> = mutableMapOf(),
 ) {
+    @get:JsonIgnore
+    public val properties: MutableMap<String, Any> = mutableMapOf()
+
     @JsonAnyGetter
     public fun `get`(): Map<String, Any> = properties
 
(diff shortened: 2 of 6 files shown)


cjbooms:
I believe we moved the instantiation of the properties map into the constructor so it would be used as part of the equals and hashcode functions. Originally it was a member instance. 

What exactly is the issue with the current approach, I have not encountered it.

networkException:
I wrote a simple test application in https://github.com/networkException/fabrikt-test

When I run the code I get the following error:

```
Exception in thread "main" com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `com.typicode.jsonplaceholder.models.TodoResponse`: Argument #2 of constructor [constructor for `com.typicode.jsonplaceholder.models.TodoResponse` (3 args), annotations: [null] has no property name annotation; must have name when multiple-parameter constructor annotated as Creator
 at [Source: (String)"{
  "userId": 1,
  "id": 1,
  "title": "delectus aut autem",
  "completed": false
}"; line: 1, column: 1]
```

cjbooms:
> I wrote a simple test application in https://github.com/networkException/fabrikt-test
> 
> When I run the code I get the following error:
> 
> ```
> Exception in thread "main" com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `com.typicode.jsonplaceholder.models.TodoResponse`: Argument #2 of constructor [constructor for `com.typicode.jsonplaceholder.models.TodoResponse` (3 args), annotations: [null] has no property name annotation; must have name when multiple-parameter constructor annotated as Creator
>  at [Source: (String)"{
>   "userId": 1,
>   "id": 1,
>   "title": "delectus aut autem",
>   "completed": false
> }"; line: 1, column: 1]
> ```
> 
> for this generated `TodoResponse.kt`:
> 
> ```kotlin
> package com.typicode.jsonplaceholder.models
> 
> import com.fasterxml.jackson.`annotation`.JsonAnyGetter
> import com.fasterxml.jackson.`annotation`.JsonAnySetter
> import com.fasterxml.jackson.`annotation`.JsonIgnore
> import com.fasterxml.jackson.`annotation`.JsonProperty
> import java.math.BigDecimal
> import javax.validation.constraints.NotNull
> import kotlin.Any
> import kotlin.String
> import kotlin.collections.Map
> import kotlin.collections.MutableMap
> 
> public data class TodoResponse(
>   @param:JsonProperty("id")
>   @get:JsonProperty("id")
>   @get:NotNull
>   public val id: BigDecimal,
>   @param:JsonProperty("userId")
>   @get:JsonProperty("userId")
>   @get:NotNull
>   public val userId: BigDecimal,
>   @get:JsonIgnore
>   public val properties: MutableMap<String, Any> = mutableMapOf(),
> ) {
>   @JsonAnyGetter
>   public fun `get`(): Map<String, Any> = properties
> 
>   @JsonAnySetter
>   public fun `set`(name: String, `value`: Any) {
>     properties[name] = value
>   }
> }
> ```

had a look at your example project but it doesn't compile for me. Looks like you might be depending on some third-party mocking layer I don't have access to

cjbooms:
It appears as though you just need to register the Kotlin module to make it work.
```
class TestAdditionalProperties {

    val objectMapper = ObjectMapper().registerKotlinModule()
    val json = """
        {
          "userId": 1,
          "id": 1,
          "title": "delectus aut autem",
          "completed": false
        }
    """.trimIndent()

    @Test
    fun testSerialization() {
        val result = objectMapper.readValue(json, TodoResponse::class.java)

        assertNotNull(result)
    }
}

```

cjbooms:
Closed as this would break the equals and hashcode generation when additional properties are used.

networkException:
> It appears as though you just need to register the Kotlin module to make it work.

Great thanks, I wasn't aware of that

PR #260, opened 2024-01-22
Support allOf with a single object in combination with nullable

networkException:
This pull request adds support for OpenAPI 3.0 style nullable references:

```yaml
foo:
  nullable: true
  allOf:
  - $ref: '#/components/schemas/Foo'
```

by special casing `allOf` with one entry and respecting either the allOfSchema or the "parent" schema being nullable.


cjbooms:
Copied to #261 so I can inspect what you are trying to achieve

PR #261, opened 2024-01-24
Support allOf with a single object in combination with nullable

cjbooms:
This patch adds support for OpenAPI 3.0 style nullable references:

```yaml
foo:
  nullable: true
  allOf:
  - $ref: '#/components/schemas/Foo'
```

by special casing `allOf` with one entry and respecting either the allOfSchema or the "parent" schema being nullable.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
@@ -127,7 +127,7 @@ class JacksonModelGenerator(
 
                 else -> className
             }
-            return if (isNullable) typeName.copy(nullable = true) else typeName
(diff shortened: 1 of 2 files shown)


cjbooms:
@networkException from my understanding, the nullable field would be declaring the foo itself can null. 

For example, for the following OAS spec
```
MyObject:
  type: object
  required:
    - foo
  properties:
    foo:
      nullable: true
      allOf:
       - $ref: '#/components/schemas/Foo'
```
below would be valid
```
{"foo": null} 
```

I'm not sure if Jackson can be configured to differentiate between absent and null, where foo in the above example can be present, null, but never absent

cjbooms:
👎

networkException:
I understand the concern and I don't know how this could be designed, I'll try (I'm struggling a bit to find my way around the codebase, but should be doable) to work something out of you can see a path forward for supporting this.

Are the changes in #265 related? I can't quite parse their effect

Issue #262, opened 2024-01-27
JSDelivr OpenAPI spec fails


avwie:
I am trying to generate https://data.jsdelivr.com/v1/spec.yaml, however it fails for Fabrikt on:

```
Exception in thread "main" java.lang.IllegalArgumentException: The Kaizen openapi-parser library threw a NPE exception when parsing this API. This is commonly due to an external schema reference that is unresolvable, possibly due to a lack of internet connection
	at com.cjbooms.fabrikt.util.YamlUtils.parseOpenApi(YamlUtils.kt:41)
```

However, using the OpenApiTools Kotlin generator it works.  I am unable to find any external schema reference in the code.

cjbooms:
Fes does not support this kind of definition:
```
            version:
              type:
                - string
                - 'null'
```

avwie:
Okay. How could I’ve found that out? And, is it documented somewhere?

cjbooms:
Fabrikt was only built to support OpenApi 3.0
The OpenApi 3.1 specification made some breaking changes to 3.0, this is one of them. 
I have created this little hack to fix it:
https://github.com/cjbooms/fabrikt/pull/263

cjbooms:
I have created two PRs which improve support for this API. However, there is a pretty crazy amount of inlining of schema definitions, and the synthetic schema names that Fabrikt is generating are conflicting with each other. 

When fabrikt sees an inlined object, it uses the top-level enclosing schema as the prefix, and the property name as the suffix when generating the data classes. Example:
```
    Dog:
      type: object
      properties:
        walker:
          allOf:
            - $ref: "#/components/schemas/Person"
            - $ref: "#/components/schemas/Company"
```
Will result in a synthetic schema name of `DogWalker`. If the Dog object has another nested properties, called walker, that also defines an inlined object, it will result in the same synthetic `DogWalker` schema name

cjbooms:
I'm closing this issue as I have fixed what issues I could. To enable generation, the JsDelivr API will need to be restructured to extract some of the inlined schemas into top-level schemas.
This can be achieved without changing the contract of the API

PR #264, opened 2024-01-28
Add support for generating aggregated inlined objects

cjbooms:
Fabrikt now supports inlined definitions of aggregated objects. For example:
```
    Dog:
      type: object
      properties:
        walker:
          allOf:
            - $ref: "#/components/schemas/Person"
            - $ref: "#/components/schemas/Company"

```
Support is preserved for the special case where the aggregation is over a single schema.
Highlighted by the API supplied in Issue #262


src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
```diff
@@ -222,26 +224,6 @@ object KaizenParserExtensions {
     fun Schema.isOneOfSuperInterface() =
         discriminator != null && discriminator.propertyName != null && oneOfSchemas.isNotEmpty()
 
-    private fun Schema.isPropertyWithAllOfSingleType() =
```
cjbooms, line 225: @acanda  It appears as though your contribution below was attempting to work around the lack of support. Hopefully this PR improves things further

acanda: The point of my PR #247 was to treat `allOf` schemas with a single `$ref` (and nothing else that changes the type) as non-aggregated schemas and keep the type name of the referenced type, i.e.
```yaml
    Dog:
      type: object
      properties:
        owner:
          description: "The registered owner of the dog."
          allOf:
            - $ref: "#/components/schemas/Person"
        walker:
          description: "The person walking the dog."
          allOf:
            - $ref: "#/components/schemas/Person"
```
results in 
```kotlin
public data class Dog(
    @param:JsonProperty("owner")
    @get:JsonProperty("owner")
    @get:Valid
    public val owner: Person? = null,  // not type DogOwner
    @param:JsonProperty("owner")
    @get:JsonProperty("owner")
    @get:Valid
    public val walker: Person? = null,  // not type DogWalker
)
```

From looking a the test case, this PR reverts those changes again.

cjbooms: Prior to your change, there was no support for inlined aggregated schemas. Your change added support for inlined allOf schemas with exactly one reference. This PR expands support for inlined schemas to include allOf anyOf with any number of referenced schemas.
The sacrifice is that if there is only in the allOf definition, it no longer short-cuts to reference that directly. But, I would say that the generated data classes are a more accurate representation of the OpenApi modelling this way

acanda: Technically you are right, aggregations, even with only one reference, should generate a new type. But OpenAPI 3.0 has one major drawback: it [ignores any sibling elements of `$ref`](https://swagger.io/docs/specification/using-ref/#sibling), so you can't directly document a property with a `$ref`. The following is unfortunately not a valid 3.0 spec:
```yaml
    Dog:
      type: object
      properties:
        owner:
          description: "The owner of the dog."
          $ref: "#/components/schemas/Person"
        walker:
          description: "The person walking the dog."
          $ref: "#/components/schemas/Person"
```
People who want to document the properties in their API have to use the workaround with `description`, `allOf` and a single `$ref`.

In my case I have to use a third party spec that uses this pattern a lot. There a re a lot of animal to person relations that need to be documented as the property name is not enough to explain the relationship. Generating the code for this spec would end in a lot of DogOwner, DogWalker, CatGroomer, PonyRider, AligatorWrangler, LionTamer, DolphinTrainer, and DinosaurResearcher classes that should just be a single Person class.

Having to deal with all those synthetic classes would make life for people in my shoes considerably harder. I'd probably have to preprocess the spec to remove all the `description` and `allOf` elements so I end up with only one Person class.

In the end it's your call. I hope Fabrikt keeps supporting a (I think quite common) pattern people use to work around a limitation of OpenAPI 3.0.

cjbooms: @acanda if you think it is beneficial for fabrkit to have the special single-value case that you original added, we can still support that:
https://github.com/cjbooms/fabrikt/pull/266/files

Let me know what you decide, I am happy with or without it. It's not too much extra complexity

acanda: I think we just posted at the same time 😄. As stated above, I'd be happier with support for the single-value case.

cjbooms: Oh nice. That's a decent justification for the extra complexity

PR #265, opened 2024-01-28
Add support for nullable keyword.

cjbooms:
If a required property has its schema flagged as nullable, it is no longer flagged as a non null.

### DIFF
--- a/src/test/resources/examples/jsonMergePatch/models/Models.kt
+++ b/src/test/resources/examples/jsonMergePatch/models/Models.kt
@@ -91,7 +91,6 @@ public data class NullabilityCheck(
     public val notNullWithDefaultNotRequired: JsonNullable<String> = JsonNullable.of(""),
     @param:JsonProperty("nullable-with-default-not-required")
     @get:JsonProperty("nullable-with-default-not-required")
-    @get:NotNull
     public val nullableWithDefaultNotRequired: JsonNullable<String?> = JsonNullable.of(""),
     @param:JsonProperty("not-null-no-default-required")
     @get:JsonProperty("not-null-no-default-required")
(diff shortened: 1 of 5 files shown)


src/test/resources/examples/jsonMergePatch/models/Models.kt
```diff
@@ -91,24 +91,21 @@ public data class NullabilityCheck(
     public val notNullWithDefaultNotRequired: JsonNullable<String> = JsonNullable.of(""),
     @param:JsonProperty("nullable-with-default-not-required")
     @get:JsonProperty("nullable-with-default-not-required")
-    @get:NotNull
```
cjbooms, line 94: @atollk or @larsduelfer could one of you please eyeball this file and let me know if it is still correct. 
I am proposing that schema's marked `nullable: true` should be nullable in the kotlin data classes, even if they are marked as required in the enclosing schema. EG:
```
    NullabilityCheck:
      type: object
      x-json-merge-patch: true
      required:
        - not-null-no-default-required
        - nullable-no-default-required
        - not-null-with-default-required
        - nullable-with-default-required
      properties:
        not-null-no-default-not-required:
          type: string
          nullable: false
        nullable-no-default-not-required:
          type: string
          nullable: true
```
However, I do not know how that plays with the merge patch stuff you contributed.

averabaq (approved):
LGTM! 👍

src/test/resources/examples/jsonMergePatch/models/Models.kt
atollk: Sorry, I wasn't too active on GH the last weeks, so I missed your question. I'm not part of the project which uses the Merge Patch features anymore, so I cannot attest to whether your change breaks something there or not, but looking at the changes here in a vacuum, they seem absolutely reasonable to me, so I'd expect it should be fine. :+1:

larsduelfer: HI @cjbooms I must have missed the notification for your mentions, apologies! This seems correct to me, too.
I am not sure if we ever mentioned it but we have forked this repo and are using our fork since a while where we did quite a lot of modification for our purposes that we have not yet managed to try to generalize and contribute back. Happy to exchange ideas for this, too.

cjbooms: @larsduelfer Any contributions you or your team have are welcome. Will facilitate in any way I can, and always opej to new maintainers to this lib

Issue #258, opened 2024-01-15
When using sealed classes: Discriminator not included in serialized object


ulrikandersen:
Hi,

When using fabrikt with `SEALED_INTERFACES_FOR_ONE_OF` I am experiencing issues with the serialised object: It does not contain the discriminator property.

The issue exists in when using fabrikt in my own code, but I have also tried using the models from [the example](https://github.com/cjbooms/fabrikt/tree/master/src/test/resources/examples/discriminatedOneOf) and see the same issues:

* The generated JSON does not contain `status` to identify the sub class.
* When trying to parse JSON that does contains the `status` it fails parsing.

I am using Jackson with no extra configuration.

### Example:
```kotlin
class JacksonParsingTest {
    @Test
    fun writeAndRead() {
        // write
        val obj = SomeObj(state = StateA)
        val objAsString = ObjectMapper().writeValueAsString(obj)
        println("Serialized object: $objAsString")

        // read
        val parsedObj = ObjectMapper().readValue("""
            {
              "state": {
                "status": "a"
              }
            }
        """.trimIndent(), SomeObj::class.java)
        println("Deserialized object: $parsedObj")
    }

    public data class SomeObj(
        @param:JsonProperty("state")
        @get:JsonProperty("state")
        @get:NotNull
        @get:Valid
        public val state: State,
    )

    @JsonTypeInfo(
        use = JsonTypeInfo.Id.NAME,
        include = JsonTypeInfo.As.EXISTING_PROPERTY,
        property = "status",
        visible = true,
    )
    @JsonSubTypes(
        JsonSubTypes.Type(value = StateA::class, name = "a"),
        JsonSubTypes.Type(
            value =
            StateB::class,
            name = "b",
        ),
    )
    public sealed interface State

    public object StateA : State

    public object StateB : State

    public enum class Status(
        @JsonValue
        public val `value`: String,
    ) {
        A("a"),
        B("b"),
        ;

        public companion object {
            private val mapping: Map<String, Status> = values().associateBy(Status::value)

            public fun fromValue(`value`: String): Status? = mapping[value]
        }
    }
}
```

Output:
```
Serialized object: {"state":{}}

com.fasterxml.jackson.databind.exc.InvalidTypeIdException: Could not resolve subtype of [simple type, class JacksonParsingTest$State]: missing type id property 'status' (for POJO property 'state')
[...]

PR #268, opened 2024-03-01
Do not strip discriminator property in oneOf generation.

cjbooms:
As per title and Issue #258 it does not seem correct to be stripping a declared property from the data classes.

Instead, adopt the same approach as allOf polymorphism and default the discriminator to the correct value

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
@@ -798,14 +798,9 @@ class JacksonModelGenerator(
     private fun List<SchemaInfo>.filterByExternalRefResolutionMode(
         externalReferences: Map.Entry<String, MutableSet<String>>,
     ) = when (externalRefResolutionMode) {
-        ExternalReferencesResolutionMode.TARGETED -> this.filter { apiSchema ->
-            externalReferences.value.contains(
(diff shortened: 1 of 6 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
```diff
@@ -462,22 +465,8 @@ class JacksonModelGenerator(
         extensions: Map<String, Any>,
         oneOfInterfaces: Set<SchemaInfo>,
     ): TypeSpec {
-        val filteredProperties = if (oneOfInterfaces.size == 1) {
```
cjbooms, line 465: This is the logic in question. I'm not sure of the intent here, but perhaps @pschichtel can weigh in before I revert it

pschichtel:
I personally think this is a mistake.

src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt
pschichtel: the logic is:

if:
  - the super interface' model defines a discriminator field
  - the discriminator mappings map to the current model exactly once
then:
  - remove the discriminator field, since it must always be the same value given by the jackson type info field.

averabaq (approved):
👍

PR #269, opened 2024-03-05
Fix inline schema definitions inherited from a parent.

cjbooms:
When an inline schema is inherited, the child should reference the parent definition, and not build ain incompatible child version


averabaq:
👍

PR #270, opened 2024-03-06
Issue with default value for additional property (oneOf)

ulrikandersen:
After #268 was merged it seems like additional properties in the generated model has an incorrect default value.

The test for discriminatedOneOf illustrates the issue: StateBMode has it's default value set to a non existent enum value - which looks like the value of the discriminator field.

Screenshot 2024-03-06 at 14 42 29

~~I would love to assist with a fix, but I am simply not familiar enough with the codebase to be able to identify where to start.~~

I have taken my best shot at fixing the issue. Let me know what you think.

I notice that the ordering of properties in the model is different from the order in the spec, but perhaps that's intentional?


cjbooms:
Nice work! 👍

PR #273, opened 2024-04-11
Support adding marker interface to oneOfs.

cjbooms:
While the discriminator property is needed to enable full polymorphism with Jackson, it should not be a blocker to  applying the sealed interface to classes that deserve it.

This PR allows the interface to be applied, without the Jackson polymorphism annotations:

```
public sealed interface State

public data class StateA(
    @get:JsonProperty("status")
    @get:NotNull
    public val status: String,
) : State

public data class StateB(
    @get:JsonProperty("mode")
    @get:NotNull
    public val mode: String,
) : State
```
Jackson won't be able to figure out which object to deserialise to, but it is beneficial to have the marker interface in code

### DIFF
--- a/src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
+++ b/src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
@@ -7,14 +7,12 @@ import kotlin.String
 public sealed interface State
 
 public data class StateA(
-    @param:JsonProperty("status")
     @get:JsonProperty("status")
     @get:NotNull
     public val status: String,
 ) : State
 
 public data class StateB(
-    @param:JsonProperty("mode")
     @get:JsonProperty("mode")
     @get:NotNull
     public val mode: String,
--- a/src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
+++ b/src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
@@ -1,0 +1,21 @@
+package examples.oneOfMarkerInterface.models
+
+import com.fasterxml.jackson.`annotation`.JsonProperty
+import javax.validation.constraints.NotNull
+import kotlin.String
+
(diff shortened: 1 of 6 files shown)


cjbooms:
@pschichtel I have started using the `SEALED_INTERFACES_FOR_ONE_OF` feature you added in some work projects recently, and I have to say I am a big fan. 💯 

Possibly I should make it the default behaviour in a version soon.

This PR extends support even more for a different use case I have.

src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.fasterxml.jackson.`annotation`.JsonProperty
+import javax.validation.Valid
+import javax.validation.constraints.NotNull
+import kotlin.String
+
+public data class Container(
+    @param:JsonProperty("state")
+    @get:JsonProperty("state")
+    @get:Valid
+    public val state: State? = null,
```
cjbooms, line 12: Here I am not sure whether I it reference State when the Polymorphic annotations are missing, or whether I should have it set to Any to ensure deserialization is not broken

averabaq: Nor I. Well, the model generated is correct but having a model that fails at deserialization defeats the purpose  🤔  I'd slightly lean for `Any`, but I'm on the fence.

averabaq (approved):
Gotcha! 👍

src/test/resources/examples/oneOfMarkerInterface/models/Models.kt
cjbooms: OK, changed to Any. That at least preserves the existing behaviour better. If anyone comes looking for it, we can add a new configuration option to allow them to opt-in

PR #278, opened 2024-04-16
Add support for skipping validation annotations

vasuman:
Trying to get this generator to work on Android and needed a way to turn off the validation annotations since `javax.validation` and `jakarta.validation` aren't supported. Could also add some validations from the `androidx.annotation` [package](https://developer.android.com/reference/androidx/annotation/package-summary) instead, if you'd prefer that approach.

Let me know if this seems acceptable


cjbooms:
LGTM, I'll try and cut you a new release later today

Issue #214, opened 2023-05-17
Ktor server controllers


atollk:
I haven't completely fleshed out this idea but just playing around with it, I feel like it shouldn't be too hard to add generation of a Ktor-based server as an option. And since afaik there are no Ktor-generators for OpenAPI in existence yet, it would probably be a pretty useful feature.

The idea would be to generate the same controller interfaces as for Spring or Micronaut right now, with the main difference that they don't use annotations but rather the Ktor DSL.

For example, Spring:

```kotlin
@Controller
@Validated
@RequestMapping("")
interface InternalEventsController {
    /**
     * Generate change events for a list of entities
     *
     * @param bulkEntityDetails
     */
    @RequestMapping(
        value = ["/internal/events"],
        produces = ["application/json", "application/problem+json"],
        method = [RequestMethod.POST],
        consumes = ["application/json"]
    )
    fun post(
        @RequestBody @Valid
        bulkEntityDetails: BulkEntityDetails
    ): ResponseEntity<EventResults>
}
```

Ktor:

```kotlin
interface InternalEventsController {
    /**
     * Generate change events for a list of entities
     *
     * @param bulkEntityDetails
     */
    fun post(
        bulkEntityDetails: BulkEntityDetails
    ): HttpResponse<EventResults>

    companion object {
        // needs to be called from somewhere
        fun addRouting(routing: Routing, controller: InternalEventsController) {
            routing.route("/internal/events") {
                post {
                    val request = call.receive<BulkEntityDetails>()
                    val response = controller.post(request)
                    call.respond(response)
                }
            }
        }
    }
}
```

atollk:
Hey. I personally do not need this feature anymore right now but I could still try and give some feedback if you want me to.

ulrikandersen:
@atollk please have a look at #280 when you have time 🙏

PR #280, opened 2024-04-18
Ktor server routes and controller interfaces

ulrikandersen:
### Design choices

Inspired by #214, Ktor server code is generated as a controller interface with methods for the routes along with a companion object which is used for mounting the routes on the Ktor server (see example further down).

Ktor is by design very flexible, and thus in order for us to be able to generate code we need to make certain design decisions. The approach with routing function + an interface should work well for most cases. The users are still free to configure Ktor however they like, including where they mount the routes and how the implement the controller interface.

Ktor's ApplicationCall is passed to the controller methods to allow them to perform additional logic on the incoming/outgoing request. One might desire to create an extension function for resolving the current user or respond directly with a redirect which requires having access to `call`.

Controller methods responsible for responding to the request (`call.respond()`). If no respond is sent, Ktor will respond with 404 per default.

**Example**

```kotlin
public interface InternalEventsController {

    public suspend fun post(bulkEntityDetails: BulkEntityDetails, call: ApplicationCall)

    public companion object {
        public fun Route.internalEventsRoutes(controller: InternalEventsController) {
            authenticate("oauth2") {
                post("/internal/events") {
                    val bulkEntityDetails = call.receive<BulkEntityDetails>()
                    controller.post(bulkEntityDetails, call)
                }
            }
        }
    }
}
```

**Usage example**

```kotlin
fun main() {
    embeddedServer(Netty, port = 8080) {
        routing {
            internalEventsRoutes(object : InternalEventsController {
                override suspend fun post(
                    bulkEntityDetails: BulkEntityDetails,
                    call: ApplicationCall
                ) {
                    call.respond(HttpStatusCode.OK, EventResults(emptyList()))
                }
            })
        }
    }.start(wait = true)
}
```

Here the controller interface is implemented as an object, but one could also implement is as a regular class which would make things like dependency injection and modularisation easier.

### Changes
1. Add Ktor route & controller interface generation.
2. Split controller interface generator abstract class in two; one for annotation based libraries (Spring, Micronaut) and one common for all.

### Known limitations
1. Headers are always handled as strings. Should be OK for now, but an improvement would be to also handle those in a typed way (including conversion from string).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,60 +188,60 @@ data class Responses(
 
 This section documents the available CLI parameters for controlling what gets generated. This documentation is generated using: `./gradlew printCodeGenUsage`
 
-| Parameter                     | Description                                                                                                                                                           |
-| ----------------------------- |-----------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-|   `--api-file`                | This must be a valid Open API v3 spec. All code generation will be based off this input.                                                                              |
-|   `--api-fragment`            | A partial Open API v3 fragment, to be combined with the primary API for code generation purposes.                                                                     |
-| * `--base-package`            | The base package which all code will be generated under.                                                                                                              |
-|   `--external-ref-resolution` | Specify to which degree referenced schemas from external files are included in model generation. Default: TARGETED                                                    |
-|                               | CHOOSE ONE OF:                                                                                                                                                        |
-|                               | `TARGETED` - Generate models only for directly referenced schemas in external API files.                                                                              |
-|                               | `AGGRESSIVE` - Referencing any schema in an external API file triggers generation of every external schema in that file.                                              |
-|   `--http-client-opts`        | Select the options for the http client code that you want to be generated.                                                                                            |
-|                               | CHOOSE ANY OF:                                                                                                                                                        |
-|                               | `RESILIENCE4J` - Generates a fault tolerance service for the client using the following library "io.github.resilience4j:resilience4j-all:+" (only for OkHttp clients) |
-|                               | `SUSPEND_MODIFIER` - This option adds the suspend modifier to the generated client functions (only for OpenFeign clients)                                             |
-|   `--http-client-target`      | Optionally select the target client that you want to be generated. Defaults to OK_HTTP                                                                                |
-|                               | CHOOSE ONE OF:                                                                                                                                                        |
-|                               | `OK_HTTP` - Generate OkHttp client.                                                                                                                                   |
-|                               | `OPEN_FEIGN` - Generate OpenFeign client.                                                                                                                             |
-|   `--http-controller-opts`    | Select the options for the controllers that you want to be generated.                                                                                                 |
-|                               | CHOOSE ANY OF:                                                                                                                                                        |
-|                               | `SUSPEND_MODIFIER` - This option adds the suspend modifier to the generated controller functions                                                                      |
-|                               | `AUTHENTICATION` - This option adds the authentication parameter to the generated controller functions                                                                |
-|   `--http-controller-target`  | Optionally select the target framework for the controllers that you want to be generated. Defaults to Spring Controllers                                              |
-|                               | CHOOSE ONE OF:                                                                                                                                                        |
-|                               | `SPRING` - Generate for Spring framework.                                                                                                                             |
-|                               | `MICRONAUT` - Generate for Micronaut framework.                                                                                                                       |
-|                               | `KTOR` - Generate for Ktor server.                                                                                                                                    |
-|   `--http-model-opts`         | Select the options for the http models that you want to be generated.                                                                                                 |
-|                               | CHOOSE ANY OF:                                                                                                                                                        |
-|                               | `X_EXTENSIBLE_ENUMS` - This option treats x-extensible-enums as enums                                                                                                 |
-|                               | `JAVA_SERIALIZATION` - This option adds Java Serializable interface to the generated models                                                                           |
-|                               | `QUARKUS_REFLECTION` - This option adds @RegisterForReflection to the generated models. Requires dependency "'io.quarkus:quarkus-core:+"                              |
-|                               | `MICRONAUT_INTROSPECTION` - This option adds @Introspected to the generated models. Requires dependency "'io.micronaut:micronaut-core:+"                              |
-|                               | `MICRONAUT_REFLECTION` - This option adds @ReflectiveAccess to the generated models. Requires dependency "'io.micronaut:micronaut-core:+"                             |
-|                               | `INCLUDE_COMPANION_OBJECT` - This option adds a companion object to the generated models.                                                                             |
-|                               | `SEALED_INTERFACES_FOR_ONE_OF` - This option enables the generation of interfaces for discriminated oneOf types                                                       |
-|                               | `NON_NULL_MAP_VALUES` - This option makes map values non-null. The default (since v15) and most spec compliant is make map values nullable                            |
-|   `--output-directory`        | Allows the generation dir to be overridden. Defaults to current dir                                                                                                   |
-|   `--resources-path`          | Allows the path for generated resources to be overridden. Defaults to `src/main/resources`                                                                            |
-|   `--src-path`                | Allows the path for generated source files to be overridden. Defaults to `src/main/kotlin`                                                                            |
-|   `--targets`                 | Targets are the parts of the application that you want to be generated.                                                                                               |
-|                               | CHOOSE ANY OF:                                                                                                                                                        |
-|                               | `HTTP_MODELS` - Jackson annotated data classes to represent the schema objects defined in the input.                                                                  |
-|                               | `CONTROLLERS` - Spring / Micronaut / Ktor HTTP controllers for each of the endpoints defined in the input.                                                            |
-|                               | `CLIENT` - Simple http rest client.                                                                                                                                   |
-|                               | `QUARKUS_REFLECTION_CONFIG` - This options generates the reflection-config.json file for quarkus integration projects                                                 |
-|   `--type-overrides`          | Specify non-default kotlin types for certain OAS types. For example, generate `Instant` instead of `OffsetDateTime`                                                   |
-|                               | CHOOSE ANY OF:                                                                                                                                                        |
-|                               | `DATETIME_AS_INSTANT` - Use `Instant` as the datetime type. Defaults to `OffsetDateTime`                                                                              |
-|                               | `DATETIME_AS_LOCALDATETIME` - Use `LocalDateTime` as the datetime type. Defaults to `OffsetDateTime`                                                                  |
-|   `--validation-library`      | Specify which validation library to use for annotations in generated model classes. Default: JAVAX_VALIDATION                                                         |
-|                               | CHOOSE ONE OF:                                                                                                                                                        |
-|                               | `JAVAX_VALIDATION` - Use `javax.validation` annotations in generated model classes (default)                                                                          |
-|                               | `JAKARTA_VALIDATION` - Use `jakarta.validation` annotations in generated model classes                                                                                |
-|                               | `NO_VALIDATION` - Use no validation annotations in generated model classes                                                                                            |
+| Parameter                     | Description |
+| ----------------------------- | ----------------------------- |
+|   `--api-file`                | This must be a valid Open API v3 spec. All code generation will be based off this input. |
+|   `--api-fragment`            | A partial Open API v3 fragment, to be combined with the primary API for code generation purposes. |
+| * `--base-package`            | The base package which all code will be generated under. |
+|   `--external-ref-resolution` | Specify to which degree referenced schemas from external files are included in model generation. Default: TARGETED |
+|                               | CHOOSE ONE OF: |
+|                               |   `TARGETED` - Generate models only for directly referenced schemas in external API files. |
+|                               |   `AGGRESSIVE` - Referencing any schema in an external API file triggers generation of every external schema in that file. |
+|   `--http-client-opts`        | Select the options for the http client code that you want to be generated. |
+|                               | CHOOSE ANY OF: |
+|                               |   `RESILIENCE4J` - Generates a fault tolerance service for the client using the following library "io.github.resilience4j:resilience4j-all:+" (only for OkHttp clients) |
+|                               |   `SUSPEND_MODIFIER` - This option adds the suspend modifier to the generated client functions (only for OpenFeign clients) |
+|   `--http-client-target`      | Optionally select the target client that you want to be generated. Defaults to OK_HTTP |
+|                               | CHOOSE ONE OF: |
+|                               |   `OK_HTTP` - Generate OkHttp client. |
+|                               |   `OPEN_FEIGN` - Generate OpenFeign client. |
+|   `--http-controller-opts`    | Select the options for the controllers that you want to be generated. |
+|                               | CHOOSE ANY OF: |
+|                               |   `SUSPEND_MODIFIER` - This option adds the suspend modifier to the generated controller functions |
+|                               |   `AUTHENTICATION` - This option adds the authentication parameter to the generated controller functions |
+|   `--http-controller-target`  | Optionally select the target framework for the controllers that you want to be generated. Defaults to Spring Controllers |
+|                               | CHOOSE ONE OF: |
+|                               |   `SPRING` - Generate for Spring framework. |
+|                               |   `MICRONAUT` - Generate for Micronaut framework. |
+|                               |   `KTOR` - Generate for Ktor server. |
+|   `--http-model-opts`         | Select the options for the http models that you want to be generated. |
+|                               | CHOOSE ANY OF: |
+|                               |   `X_EXTENSIBLE_ENUMS` - This option treats x-extensible-enums as enums |
+|                               |   `JAVA_SERIALIZATION` - This option adds Java Serializable interface to the generated models |
+|                               |   `QUARKUS_REFLECTION` - This option adds @RegisterForReflection to the generated models. Requires dependency "'io.quarkus:quarkus-core:+" |
+|                               |   `MICRONAUT_INTROSPECTION` - This option adds @Introspected to the generated models. Requires dependency "'io.micronaut:micronaut-core:+" |
+|                               |   `MICRONAUT_REFLECTION` - This option adds @ReflectiveAccess to the generated models. Requires dependency "'io.micronaut:micronaut-core:+" |
+|                               |   `INCLUDE_COMPANION_OBJECT` - This option adds a companion object to the generated models. |
+|                               |   `SEALED_INTERFACES_FOR_ONE_OF` - This option enables the generation of interfaces for discriminated oneOf types |
+|                               |   `NON_NULL_MAP_VALUES` - This option makes map values non-null. The default (since v15) and most spec compliant is make map values nullable |
+|   `--output-directory`        | Allows the generation dir to be overridden. Defaults to current dir |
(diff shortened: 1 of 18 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
```diff
@@ -0,0 +1,362 @@
...
+
+    /**
+     * Builds a utility object to be used for parsing query parameters
+     */
+    private fun buildRoutingUtils(): ControllerLibraryType {
+        val genericR = TypeVariableName("R", Any::class)
+
+        val returnType = genericR.copy(nullable = true, reified = true)
+
+        val spec = TypeSpec.objectBuilder("RoutingUtils")
```
atollk, line 264: I don't quite see why this should be a `public object` instead of some `private fun`ctions. Foremost, this feels like it could cause name clashes within a package if you have multiple controller files. Also, it might clutter the import space.

src/test/resources/examples/githubApi/controllers/ktor/Controllers.kt
```diff
@@ -0,0 +1,823 @@
...
+        id: String,
+        pullRequest: PullRequest,
+    ): ControllerResult<Unit>
+
+    public companion object {
+        public fun Route.repositoriesPullRequestsRoutes(controller: RepositoriesPullRequestsController) {
+            get("/repositories/{parent-id}/pull-requests") {
+                val parentId = call.parameters.getOrFail<kotlin.String>("parent-id")
+                val xFlowId = call.request.headers["X-Flow-Id"]
+                val limit = call.request.queryParameters.getTyped<kotlin.Int>("limit")
```
atollk, line 741: It seems that schema information is not used here, unlike it is for the Spring controllers. Wouldn't it make sense to add checks here as well?

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
```diff
@@ -0,0 +1,362 @@
...
+                PropertySpec.builder("status", ClassName("io.ktor.http", "HttpStatusCode"))
+                    .initializer("status")
+                    .build()
+            )
+            .addProperty(
+                PropertySpec.builder("message", genericT)
+                    .initializer("message")
+                    .build()
+            )
+            .build()
```
atollk, line 251: I think this is too limited of a response struct, considering what Ktor offers in general: https://ktor.io/docs/server-responses.html

atollk (changes requested):

atollk:
It seems like the "authentication" test case isn't extended? I think that's an important element of generation that needs to be covered.

src/test/resources/examples/binary/controllers/ktor/Controllers.kt
```diff
@@ -0,0 +1,28 @@
+package ie.zalando.controllers
```
atollk, line 1: It seems like the "authentication" test case isn't extended? I think that's an important element of generation that needs to be covered.

atollk:
I mostly looked at the examples in the test cases but from that it looks close to what I imaged :+1:

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: The two functions `Parameters.getTyped` and `Headers.getOrFail` are used by routes where parameter and header parsing is performed, and therefore contained in a utility object. 

The object is only generated once through the `generateLibrary` function.

They are compliments to the built in Ktor server function [Parameters.getOrFail](https://api.ktor.io/ktor-server/ktor-server-core/io.ktor.server.util/[jvm-and-nix]get-or-fail.html).

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: Yeah I agree it is too limited to cover all use cases, and this is why I have chosen to pass `ApplicationCall` to the controllers which allows them to both read from the request, but also respond in the various ways Ktor [supports](https://ktor.io/docs/server-responses.html).

Unfortunately this leads to an ambiguous API where the controller method is expected to return a ControllerResult<Unit>, but is also free to "close" the call with e.g. `call.respondRedirect`. 

In practise this leads to the following implementation:
```kotlin
call.respondRedirect("https://some.other.domain/path")
return ControllerResult(HttpStatusCode.OK, Unit)
```
Ktor will respond with the first `respond` and the `ControllerResult` will be ignored.

In the initial version I had simply skipped the ControllerResult in cases where the spec does not mention a return model. This would avoid the awkward code mentioned above.

Possible alternatives would be to
1. Not force a return type and let the controller call `call.respondX`. Here we loose the type safety of the return type, which is not desirable. 
2. Provide a return struct that supports all the same `respond` as Ktor. This would have to be updated whenever Ktor changes.
3. Other alternatives?

src/test/resources/examples/githubApi/controllers/ktor/Controllers.kt
ulrikandersen: I assume you mean the validation constraints added as annotations to Spring and Micronaut controllers. 

Unfortunately Ktor does not have such thing built in, and we would have to either rely on an external library or implement the checks ourselves.

I have used the [RequestValidation plugin](https://ktor.io/docs/server-request-validation.html) combined Hibernate Validator with to validate the incoming body, perhaps something similar could be done for query, path and header parameters, but it would have to be custom coded.

src/test/resources/examples/binary/controllers/ktor/Controllers.kt
ulrikandersen: Good point @atollk! 🔒 I have added the authentication test case for Ktor too.

The same commit also [adds a check](https://github.com/cjbooms/fabrikt/pull/280/commits/c82de2f89b0289d8459b7b300c5feb7b62777d2c#diff-e6e8af2766ab85eea84037212904553a47b13bcfa130967020f1a3b0ed1272d3R157) to see if `ControllerCodeGenOptionType.AUTHENTICATION` is enabled and only in that case generate auth code.

README.md
```diff
@@ -187,59 +188,60 @@ data class Responses(
 
 This section documents the available CLI parameters for controlling what gets generated. This documentation is generated using: `./gradlew printCodeGenUsage`
 
-| Parameter                     | Description |
```
cjbooms, line 190: Has this been regenerated using  `./gradlew printCodeGenUsage`?

ulrikandersen: It has now ✨🙂

cjbooms:
Let me know when this is ready for review. I should be able to cut a release for you today if you like.
I won't have any input on the generated code, as I have never used Ktor, so will leave that up to you two to decide on. It does not have to be perfect first time around, we can cut a few different releases until you are satisfied it handles your local cases

cjbooms:
LGTM - Let me know if you are happy to merge

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
atollk: > The two functions Parameters.getTyped and Headers.getOrFail are used by routes where parameter and header parsing is performed, and therefore contained in a utility object.
> The object is only generated once through the generateLibrary function.

Yeah I get the use case. I just meant that I wouldn't have used a wrapping object which afaik doesn't serve any purpose here other than syntactic structure. But I understand that might just be a stylistic preference, so I don't want to argue against it.

> They could also be private functions on the controller interfaces, but they would then be replicated across all controllers. But perhaps that's OK?

I do believe private functions would be better (i.e. private free functions, or functions within a private object). Otherwise they become part of the public interface given by Fabrikt, which we probably don't want (Hyrum's law :) )

src/test/resources/examples/githubApi/controllers/ktor/Controllers.kt
atollk: Yeah, that's what I meant. At the moment, the supported validations are numeric min/max, nullability, regex patterns and sequence length, all of which I think could be implemented in a few lines. But as @cjbooms said themselves below, not everything needs to be perfect from the start, so if you feel like this addition would be too much for this PR, I also won't mind if it gets merged without this feature. We should probably just create another issue and a note in the documentation in that case.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
atollk: That is a dilemma :)

What about creating two functions for each endpoint; a "simple" and a "powerful" one.

```kotlin

interface MyController {
    // Only returns 200 with the returned body; or 500 on an exception.
    public suspend fun post(
        querySomeObject: String,
        bodySomeObject: SomeObject,
    ): Unit
    
    // Can use arbitrary returns.
    public suspend fun post(
        call: ApplicationCall,
        querySomeObject: String,
        bodySomeObject: SomeObject,
    ) {
        val responseBody = post(querySomeObject, bodySomeObject)
        call.respond(responseBody)
        // ...
    }
    
    // Routing ...
}
```

The only issue I see is if users override both functions for one endpoint, but I feel like that would be a rare mistake and not that difficult to catch.

Of course, it is arguable whether the first method actually serves any purpose or if 99% of use cases require the second one anyway.

src/test/resources/examples/binary/controllers/ktor/Controllers.kt
atollk: Maybe I am missing something but it looks to me like it doesn't actually generate any authentication code, does it? I would expect the logic from https://ktor.io/docs/server-auth.html to be present in a generated controller, i.e. perform auth checks on relevant routes and extract the `Principal` object and pass it to the function stubs.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: Excellent point about not polluting the public API of the generated code. I have made the two functions part of the controller interface's companion object which allows them to be used by the route function without polluting the API.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: Having looked more into this, I would like to go with simply not requiring the implementation to return `ControllerResult` in cases where the route has no return content. 

In that case Ktor will respond with `200 OK` by default, but we still allow the implementation to respond with a e.g. a redirect if needed.

This change has been added in https://github.com/cjbooms/fabrikt/pull/280/commits/35c3195e9a43e1a3498a59ca5201d5078988e8a3.

src/test/resources/examples/githubApi/controllers/ktor/Controllers.kt
ulrikandersen: Yes, let's add that as an issue 👍 It would be a good improvement.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: Turns out I was wrong about the default: if no response is provided Ktor will respond with `404 Not Found` 😬 

@atollk Would you be OK with making routes without response content return `HttpStatusCode`? If the implementation choses to perform an early `call.respond` the code is a little ambiguous, but that is already the case with the endpoints that return content. Here the implementation would still be required to return a `ControllerResult` even though it performs an early respond.

```kotlin
interface Controller {
  suspend fun testPath(call: ApplicationCall): HttpStatusCode
  
  companion object {
    get("/test") {
      val resultStatus = controller.testPath(call)
      call.respond(resultStatus)
    }
  }
}

// respond with 200 OK

override suspend fun testPath(call: ApplicationCall): HttpStatusCode {
  return HttpStatusCode.OK
}

// or redirect

override suspend fun testPath(call: ApplicationCall): HttpStatusCode {
  call.respondRedirect("/someOtherPath")
  return HttpStatusCode.InternalServerError // <-- should never be reached
}
```

src/test/resources/examples/binary/controllers/ktor/Controllers.kt
atollk: You can actually just use `call.principal<Principal>()` and leave the specific cast to the user implementation. That's how I do it in my code base at least and so far I haven't encountered any issue with it.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
atollk: I don't know, I am not a fan of forcing the implementation to have a return even if will never be reached; that'd be akin to forcing users to have a `return Unit` in all their functions which don't actually return anything.

I think that a function without any return that forces the user to implement any responses using the `ApplicationCall` object directly is the best way. It's the only option that provides generic support for all use cases and that's the option that should be included in a tool like Fabrikt imo. Special convenience options could be added in later revisions in theory.

ulrikandersen: I agree that it is not nice to force a return that will never be reached.

I like the design where we provide a simple and an advanced option, and I think it could cover all use cases. Any time the user wishes to deal directly with `call` (either to read additional information from the request, or to provide a custom response) he would implement the advanced function.

I am thinking it could look like this, where the default implementation of the advanced function can be overriden by the user:

```kotlin
public interface InternalEventsController {

  // simple version
  public suspend fun post(bulkEntityDetails: BulkEntityDetails): ControllerResult<EventResults>

  // advanced version; user is responsible for returning a response
  public suspend fun post(bulkEntityDetails: BulkEntityDetails, call: ApplicationCall) {
    // default implementation delegates to the simple version
    val result = post(bulkEntityDetails)
    call.respond(result.status, result.message)
  }

  public companion object {
    public fun Route.internalEventsRoutes(controller: InternalEventsController) {
      post("/internal/events") {
        val bulkEntityDetails = call.receive<BulkEntityDetails>()
        controller.post(bulkEntityDetails, call)
      }
    }
}
```

In cases where no model is returned by the endpoint we could simply return HttpStatusCode from the simple function:

```kotlin
public interface InternalEventsController {

  public suspend fun post(bulkEntityDetails: BulkEntityDetails): HttpStatusCode

  public suspend fun post(bulkEntityDetails: BulkEntityDetails, call: ApplicationCall) {
    val status = post(bulkEntityDetails)
    call.respond(status)
  }

  public companion object {
    public fun Route.internalEventsRoutes(controller: InternalEventsController) {
      post("/internal/events") {
        val bulkEntityDetails = call.receive<BulkEntityDetails>()
        controller.post(bulkEntityDetails, call)
      }
    }
}
```

Does that resonate with you?

src/test/resources/examples/binary/controllers/ktor/Controllers.kt
ulrikandersen: Amazing 🙌 This has been added in https://github.com/cjbooms/fabrikt/pull/280/commits/6535f06e7d7561563c6af25f508d0b8a874594d9.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
ulrikandersen: I just realised that the above will not cut it: The user will be forced to always implement the simple version as it is abstract 🤔 

We could however provide a default implementation which throws in case it has not been overridden. This will happen in cases where neither the advanced nor the simple version has been implemented.

Downside is that the user will not get warned at compile time if they have forgotten to implement one of the routes.

I am all ears if you have alternative solutions i mind.

```kotlin
public suspend fun post(bulkEntityDetails: BulkEntityDetails): ControllerResult<EventResults> {
  throw BadRequestException("Not implemented")
}

public suspend fun post(bulkEntityDetails: BulkEntityDetails, call: ApplicationCall) {
  // default implementation delegates to the simple version
  val result = post(bulkEntityDetails)
  call.respond(result.status, result.message)
}
```

atollk: > I just realised that the above will not cut it: The user will be forced to always implement the simple version as it is abstract 🤔

True, I didn't think about that either. 

The `NotImplementedError` is not great, as you said, because it's lacking compile checks but that would also have been my idea at the moment. Maybe also respond a 501 status code to the call.

I'm still not to keen on the `ControllerResult` custom class. Maybe we should create a feature request on the Ktor tracker to see if Jetbrains would be willing to add a `Response`-like type that we could use. Fabrikt isn't the first time I wish something like that existed :)

But for the time being, if you want to merge the MR, it might be the right choice to just offer the "advanced" version for now and potentially add convenience overloads in the future? We could add a comment about the excepted return type to make it slightly less annoying to the user.

cjbooms: Would having an extension property help resolve this in any way? Eg, allow the user to specify the return type desired for each operation

x-ktor-response-type: blah

I'm not fully following this PR, so feel free to disregard this comment if not helpful

ulrikandersen: Let us stick with offering a function which delegates responding to the implementation (aka the advanced version) 👍 

I have added extra documentation to the generated methods which informs the user about the responsibility to respond to the call.

```kotlin
public interface InternalEventsController {
    /**
     * Generate change events for a list of entities
     *
     * Route is expected to respond with [examples.githubApi.models.EventResults].
     * Use [io.ktor.server.response.respond] to send the response.
     *
     * @param bulkEntityDetails
     * @param call The Ktor application call
     */
    public suspend fun post(bulkEntityDetails: BulkEntityDetails, call: ApplicationCall)

    public companion object {
        /**
         * Mounts all routes for the InternalEvents resource
         *
         * - POST /internal/events Generate change events for a list of entities
         */
        public fun Route.internalEventsRoutes(controller: InternalEventsController) {
            post("/internal/events") {
                val bulkEntityDetails = call.receive<BulkEntityDetails>()
                controller.post(bulkEntityDetails, call)
            }
        }
        // ...
    }
}
```

With this change the `ControllerResult` class as well as the `generateLibrary` function on the controller generator has been made obsolete, which is why I have removed them again.

Following this decision I believe the PR is ready to be merged, @cjbooms 🚀

atollk:
lgtm @cjbooms

cjbooms:
Great news. 
@ulrikandersen is the README still accurate if I use it as the squashed commit message?

ulrikandersen:
> Great news. @ulrikandersen is the PR description still accurate if I use it as the squashed commit message?

Yes, I would say so! I updated it following the decision on how to handle returns.

cjbooms:
Cutting you a release now

Issue #281, opened 2024-04-25
Duplicate class definition in nested polymorphic models


y-marion:
# Issue Description
I have tried to use nested polymorphic models as described and found an issue, that arises when generating the files.
If a nesting (beyond one level) is present, the classes of the n-th level (which still have children) is always included in the class file of it's level, but also of the (n-1)-th level.
That in turn means that the code won't compile because of multiple class definitions.

# Expected behavior

No duplicate classes are generated; each superclass is only generated in it's own class-file.

# Root cause
The issue stems from the [`groupClasses` function](https://github.com/cjbooms/fabrikt/blob/cbb86bdc4af73e9252c30470012fc18eb3721971/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt#L66) as far as i understand it.
This function associate models, which are sealed with classes that have this model as a superclass. The association does not in turn check if the class that it's trying to add is a sealed class itself. I would propose to fix this by filtering the associated models so that it's made sure that these are not sealed classes. See #283 for the tests that show this issue & #284 for this proposed fix.

## Example of issue
As example, you can try generating the api from [nestedPolymorphicModels/api.yaml](https://github.com/cjbooms/fabrikt/blob/master/src/test/resources/examples/nestedPolymorphicModels/api.yaml) via the cli.

This results in the files `CommonObject`, `FirstLevelChild`, `FirstLevelDiscriminator`, `RootDiscriminator`, `RootType`, `SecondLevelChild1`, `SecondLevelChild2`, `SecondLevelDiscriminator`, `SecondLevelMetadata`.

The errors are present in `RootType`, `FirstLevelChild`, `SecondLevelChild1`, `SecondLevelChild2` (only attaching RootType & FirstLevelChild).

RootType.kt
```kotlin
@JsonTypeInfo(
  use = JsonTypeInfo.Id.NAME,
  include = JsonTypeInfo.As.EXISTING_PROPERTY,
  property = "rootDiscriminator",
  visible = true,
)
@JsonSubTypes(JsonSubTypes.Type(value = FirstLevelChild::class, name = "firstLevelChild"))
public sealed class RootType(
  public open val rootField1: String,
  public open val rootField2: Boolean? = null,
) {
  public abstract val rootDiscriminator: RootDiscriminator
}

@JsonTypeInfo(
  use = JsonTypeInfo.Id.NAME,
  include = JsonTypeInfo.As.EXISTING_PROPERTY,
[...]

PR #282, opened 2024-04-25
Fix: Remove duplicate classes in nested polymorphic models

y-marion:
See issue #281 for detailed explanation

this also adds a test which compares the created model files (individual model files) to the generated files, to check for differences.
the changes in KotlinGenModels doesn't introduce changes in any other api files apart from the nestedPolymorphicModel file

Sorry for creating this many files, i wanted to make sure my changes didn't cause any regressions in the model generation of other api files


y-marion:
Perhaps the test for models with the single file output could be deleted with this new test

cjbooms:
> Perhaps the test for models with the single file output could be deleted with this new test

Yes, I think this is a good idea. Let's do that as a precursor PR so we can both see the problem you have found and fix you have provided. Nice work

y-marion:
closing this in favor of #283 & #284 @cjbooms

PR #284, opened 2024-04-25
Fix: Remove duplicate classes in nested polymorphic models

y-marion:
This is based on #283. 
As discussed this is the fix for the issue that I'm seeing in the nested polymorphic model creation as separate pull request.

See issue #281 for detailed explanation


cjbooms:
Moved to #285

PR #285, opened 2024-04-25
fix: don't duplicate classes in nested polymorphic models

cjbooms:
Removing code that grouped Sealed classes into a Single file. This was causing issues with multi-level sealed classes as reported in #281 
Since Kotlin 1.5, there is no need to group classes implementing a Sealed class into a single file. 
https://kotlinlang.org/docs/whatsnew15.html#package-wide-sealed-class-hierarchies


src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt
```diff
@@ -67,7 +67,9 @@ private fun <T : GeneratedType> groupClasses(allModels: Collection<T>): Map<T, L
     val sealedClasses = allModels
```
cjbooms, line 67: @ulrikandersen / @averabaq Do you think we should just remove this grouping? Grouping Sealed class implementations hasn't been a requirement since Kotlin 1.5

y-marion:
> @y-marion Let me know what you think

This solution is fine for me. I wasn't sure what the idea behind the grouping was, so I wanted to take the path with the least amount of change, which is why I decided to filter it.

src/main/kotlin/com/cjbooms/fabrikt/model/KotlinGenModels.kt
averabaq: Sorry for jumping late into this. By all means, my vote is an absolute "yes". There shouldn't be many adopters out there still on kotlin < 1.5, and if there are, they can still use Fabrikt up to this version. Still, not a big deal upgrading

PR #286, opened 2024-04-26
Ktor typed respond functions

ulrikandersen:
In #280 it was discussed how to provide the user with a type safe way to respond to the request.

It was [decided](https://github.com/cjbooms/fabrikt/pull/280#discussion_r1579041527) to fully delegate the responsibility of responding to the controller implementation, which provides the user with direct access to `call` and maintains full flexibility.

In order to be able to offer a type safe way to respond, this PR introduces two new functions on `call`:
- `call.repondTyped(status: HttpStatusCode, message: T)`
- `call.repondTyped(message: T)`

The functions are provided through a decorator on ApplicationCall (which is luckily an interface) and can be used as a drop-in replacement for ApplicationCall.

The `TypedApplicationCall` class is included as a utility class through a `generateLibrary` functionality inspired by the [client generators](https://github.com/cjbooms/fabrikt/blob/master/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGenerator.kt#L9).

Routes that do not respond with content is provided with the regular ApplicationCall as there is no need for type safety there.

Offering this type safety makes the experience much more enjoyable and removes the mental load of remembering to check which type should be returned 🧘 

https://github.com/cjbooms/fabrikt/assets/11440456/1b2b2d19-2222-48b7-82f7-bf2c45fc1059

**Examples:**

Generated class: [TypedApplicationCall](https://github.com/ulrikandersen/fabrikt/blob/3274ee9b4b83c7954a1df765f37b18c4e978ad4c/src/test/resources/examples/authentication/controllers/ktor/Controllers.kt#L339)

Route with response content: [POST /internal/events](https://github.com/ulrikandersen/fabrikt/blob/3274ee9b4b83c7954a1df765f37b18c4e978ad4c/src/test/resources/examples/githubApi/controllers/ktor/Controllers.kt#L48)

Route without response content: [GET /required](https://github.com/ulrikandersen/fabrikt/blob/3274ee9b4b83c7954a1df765f37b18c4e978ad4c/src/test/resources/examples/authentication/controllers/ktor/Controllers.kt#L32)

### DIFF
--- a/src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
+++ b/src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
@@ -34,7 +34,7 @@ public interface TestController {
          */
         public fun Route.testRoutes(controller: TestController) {
             `get`("/test") {
-                controller.test(TypedApplicationCall.from(call))
+                controller.test(TypedApplicationCall(call))
             }
         }
 
@@ -79,7 +79,7 @@ public interface TestController {
  *
  * @param R The type of the response body
  */
-public class TypedApplicationCall<R : Any> private constructor(
+public class TypedApplicationCall<R : Any>(
     private val applicationCall: ApplicationCall,
 ) : ApplicationCall by applicationCall {
     @Suppress("unused")
@@ -90,10 +90,5 @@ public class TypedApplicationCall<R : Any> private constructor(
     @Suppress("unused")
     public suspend inline fun <reified T : R> respondTyped(status: HttpStatusCode, message: T) {
         respond(status, message)
-    }
-
-    public companion object {
-        public fun <R : Any> from(applicationCall: ApplicationCall): TypedApplicationCall<R> =
-            TypedApplicationCall<R>(applicationCall)
     }
 }
--- a/src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
+++ b/src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
@@ -1,25 +1,30 @@
 package examples.singleAllOf.controllers
 
+import examples.singleAllOf.models.Result
 import io.ktor.http.Headers
+import io.ktor.http.HttpStatusCode
 import io.ktor.http.Parameters
 import io.ktor.server.application.ApplicationCall
+import io.ktor.server.application.call
 import io.ktor.server.plugins.BadRequestException
 import io.ktor.server.plugins.ParameterConversionException
+import io.ktor.server.response.respond
 import io.ktor.server.routing.Route
 import io.ktor.server.routing.`get`
 import io.ktor.util.converters.DefaultConversionService
 import io.ktor.util.reflect.typeInfo
 import kotlin.Any
 import kotlin.String
+import kotlin.Suppress
 
 public interface TestController {
     /**
      * Route is expected to respond with [examples.singleAllOf.models.Result].
-     * Use [io.ktor.server.response.respond] to send the response.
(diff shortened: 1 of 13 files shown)


src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
```diff
@@ -66,3 +71,29 @@ public interface TestController {
...
+
+    @Suppress("unused")
+    public suspend inline fun <reified T : R> respondTyped(status: HttpStatusCode, message: T) {
+        respond(status, message)
+    }
+
+    public companion object {
+        public fun <R : Any> from(applicationCall: ApplicationCall): TypedApplicationCall<R> =
+            TypedApplicationCall<R>(applicationCall)
+    }
```
atollk, line 98: I like the general idea. I haven't looked at the implementation yet or thought about potential pitfalls. For the moment I'm mostly wondering: why a private constructor with a factory function if the factory function is the exact same as the constructor, just with a different name?

src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
```diff
@@ -66,3 +71,29 @@ public interface TestController {
...
+ *
+ * @param R The type of the response body
+ */
+public class TypedApplicationCall<R : Any> private constructor(
+    private val applicationCall: ApplicationCall,
+) : ApplicationCall by applicationCall {
+    @Suppress("unused")
+    public suspend inline fun <reified T : R> respondTyped(message: T) {
+        respond(message)
+    }
```
atollk, line 88: Why the `T` generic and not just `message: R`?

src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
ulrikandersen: The factory function is indeed not needed 👍 I have removed it.

src/test/resources/examples/singleAllOf/controllers/ktor/Controllers.kt
ulrikandersen: I am not 100% why, but this setup is how I managed to make it work with the original call's `respond`:

Screenshot 2024-04-28 at 13 42 20

But maybe I am missing something?

We can also use the `respond` function variants that take a TypeInfo (https://api.ktor.io/ktor-server/ktor-server-core/io.ktor.server.response/respond.html):

```kotlin
public class TypedApplicationCall<R : Any> (
  private val applicationCall: ApplicationCall,
) : ApplicationCall by applicationCall {
  @Suppress("unused")
  public suspend inline fun <reified R> respondTyped(message: R) {
    respond(message, typeInfo<R>())
  }

  @Suppress("unused")
  public suspend inline fun <reified R> respondTyped(status: HttpStatusCode, message: R) {
    respond(status, message, typeInfo<R>())
  }
}
```

Let me know what you think is best.

atollk: I guess it doesn't really matter because either way, the generated `Response` class is final (or rather, not `open`), so the generic argument will always be the same type anyway, right?

ulrikandersen: Yes, the models are final, so `R` will always be the model and not a subclass.

cjbooms:
Everyone happy here?

ulrikandersen:
> Everyone happy here?

Sound like we agree on the API being good and an improvement compared to the type-less `call.respond`.

From my point of view it is ready to merge. We can always change the implementation if @atollk or someone else has suggestions for improvements ✌️

PR #291, opened 2024-06-20
fix: Nullable object fields now working as expected

### LINKED ISSUE #289: Required but nullable object field is non-nullable in generated class
When running a schema like this through Fabrikt:
```yaml
components:
  schemas:
    MyObject:
      type: object
      required:
        - foo
      properties:
        foo:
          type: object
          nullable: true
```
The generated class looks like this:
```kotlin
public data class MyObject(
  @param:JsonProperty("foo")
  @get:JsonProperty("foo")
  @get:NotNull
  public val foo: Map<String, Any>,
)
```
But if I'm not mistaken, one would expect:
```kotlin
public data class MyObject(
  @param:JsonProperty("foo")
  @get:JsonProperty("foo")
  @get:NotNull
  public val foo: Map<String, Any?>,
)
```

DavidMazarro:
This PR attempts to provide a fix to the issue where a nullable field in an object would appear in the generated model as non-nullable.

I also tried addressing [this other issue](https://github.com/cjbooms/fabrikt/issues/289#issuecomment-2163186058) reported by @wduncanfraser but I still haven't been able to get a fix in place for that. Feel free to pick up where I left off and add the fix to this PR (or create another issue and submit a different PR for that). I'll try to give a head start to anyone that wants to give it a go with my latest findings (I could be wrong, I'm new to this library): [the beginning part of `toModelType` recursive function](https://github.com/cjbooms/fabrikt/blob/6b7334902f3bbb99cf8792a09b2dbda8b8701849/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt#L78-L90) is used to generate the type of an OAS array (a Kotlin list). The starting call site is in the [`addToClass` function](https://github.com/cjbooms/fabrikt/blob/6b7334902f3bbb99cf8792a09b2dbda8b8701849/src/main/kotlin/com/cjbooms/fabrikt/generators/model/JacksonModelGenerator.kt#L688-L698). This populates the `isNullable` argument with the nullability information about the array, but the nullability of the items schema gets lost in the recursion. I think solving this issue would require some change that allowed to propagate the items schema information inside `toModelType`.

Closes #289.


DavidMazarro:
@cjbooms The build is failing because of the `nullableArrayItems` test I added which still doesn't have a fix. For the sake of getting the other fix merged, I was thinking of removing the `nullableArrayItems` test from this PR and create a new one that I'll leave open to resume that work there. WDYT?

cjbooms:
Sounds good. Exactly what I was gonna do before getting swamped in work. Should have time before next Tuesday to get this merged and a release cutn

DavidMazarro:
@cjbooms Done, the model generator tests should pass now.

cjbooms (approved):

cjbooms:
Thanks @DavidMazarro. Have been trying to cut you a new release, but Sonatype has changed their deployment process and it seems to have broken. Will have another attempt later in the weekend.

cjbooms:
Released. thanks for the contribution

DavidMazarro:
@cjbooms My pleasure!

PR #292, opened 2024-06-21
Add support for InputStream

ctasada:
Fabrikt already offers support for `ByteArray`, but when dealing with large files, for example, the usage of `ByteArray` puts too much pressure in memory.

This MR provides `BYTEARRAY_AS_INPUTSTREAM`. A new `--type-overrides` that allows to provide an `InputStream` when dealing with binary data.


cjbooms:
Thanks for the contribution. It'll be Monday before I get to look at this. Unless @averabaq can handle it before then

averabaq:
This is a good one that's been on the radar for some time. I'll try to put some eyes on it over the weekend. Thank you so much for this @ctasada

averabaq:
The server side code-generation looks good to me, no objections on this. On the client side, however, the code generated is incorrect:

```
    public fun postExample(
        applicationOctetStream: InputStream,
        additionalHeaders: Map<String, String> = emptyMap(),
    ): ApiResponse<Unit> {
        ...

        val request: Request = Request.Builder()
           ....
            .post(objectMapper.writeValueAsString(applicationOctetStream).toRequestBody("application/octet-stream".toMediaType()))          
           ....
    }
```    
Supporting streaming on the client side is not that simple. This is only for `OkHttp` and we also have support for `OpenFeign`. Since this is a new feature and not a breaking change, I wouldn't mind to go ahead with this, but we should either:

1) Allow input streams on the models for now and resolve to byte-arrays on request bodies regardless.
2) Don't allow input streams on client-side generation at all (i.e. ignore the `byte_array_as_input_stream` cli option)
3) Add full stream support on client side.

I can't proceed as is, sorry. Let's see what @cjbooms thinks about it  next week.

cjbooms:
> The server side code-generation looks good to me, no objections on this. On the client side, however, the code generated is incorrect:
> 
> ```
>     public fun postExample(
>         applicationOctetStream: InputStream,
>         additionalHeaders: Map<String, String> = emptyMap(),
>     ): ApiResponse<Unit> {
>         ...
> 
>         val request: Request = Request.Builder()
>            ....
>             .post(objectMapper.writeValueAsString(applicationOctetStream).toRequestBody("application/octet-stream".toMediaType()))          
>            ....
>     }
> ```
> 
> Supporting streaming on the client side is not that simple. This is only for `OkHttp` and we also have support for `OpenFeign`. Since this is a new feature and not a breaking change, I wouldn't mind to go ahead with this, but we should either:
> 
> 1. Allow input streams on the models for now and resolve to byte-arrays on request bodies regardless.
> 2. Don't allow input streams on client-side generation at all (i.e. ignore the `byte_array_as_input_stream` cli option)
> 3. Add full stream support on client side.
> 
> I can't proceed as is, sorry. Let's see what @cjbooms thinks about it next week.

Option 2 sounds good to me. Allow it on server-side generation only

cjbooms:
What is need to get this across the line @averabaq ? Could it be merged as is with the configuration option as protection against the client side generation being incorrect?

averabaq:
> Could it be merged as is with the configuration option as protection against the client side generation being incorrect?

Yes, I didn't make it go through because I was waiting for getting a PR ready with the required changes, but haven't had the chance to work on it

PR #296, opened 2024-07-08
fix: Nullable array items model generation now working as expected

### LINKED ISSUE #295: Nullable array items model generation not working as expected
Originally reported by @wduncanfraser [in this comment](https://github.com/cjbooms/fabrikt/issues/289#issuecomment-2163186058).

When running a schema like this through Fabrikt:
```yaml
openapi: 3.0.0
components:
  schemas:
    NullableArrayItems:
      type: object
      required:
        - anArray
      properties:
        anArray:
          type: array
          nullable: true
          items:
            type: string
            nullable: true
```

The generated class currently looks like:
```kotlin
public data class NullableArrayItems(
  @param:JsonProperty("anArray")
  @get:JsonProperty("anArray")
  public val anArray: List<String>?,
)
```

But it should be:
```kotlin
public data class NullableArrayItems(
  @param:JsonProperty("anArray")
  @get:JsonProperty("anArray")
  public val anArray: List<String?>?,
)
```

DavidMazarro:
WIP. For now, this PR includes the test that needs to pass to consider the issue fixed, but we still don't have a fix ready.

Closes #295.

### DIFF
--- a/src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
+++ b/src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
@@ -64,6 +64,7 @@ class ModelGeneratorTest {
         "binary",
         "oneOfMarkerInterface",
         "requiredButNullableObjectProperty",
+        "nullableArrayItems",
     )
 
     @BeforeEach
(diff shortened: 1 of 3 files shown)


DavidMazarro:
@cjbooms I think you closed this issue as unsolved instead of as "completed".

PR #299, opened 2024-07-13
add end2end tests for jackson model generation

zettelmj:
add end2end tests for jackson model generation

### DIFF
--- a/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
+++ b/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
@@ -1,12 +1,12 @@
 package com.cjbooms.fabrikt.models.jackson
 
+import com.cjbooms.fabrikt.models.jackson.Helpers.mapper
 import com.example.models.*
-import com.fasterxml.jackson.module.kotlin.jsonMapper
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.jupiter.api.Test
 
 class DeepNestedPolymorphismTest {
-    private val objectMapper = jsonMapper()
+    private val objectMapper = mapper()
     private val writer = objectMapper.writerWithDefaultPrettyPrinter()
 
     @Test
--- a/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
+++ b/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
@@ -1,0 +1,57 @@
+package com.cjbooms.fabrikt.models.jackson
+
+import com.example.models.*
+import com.fasterxml.jackson.module.kotlin.jsonMapper
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class DeepNestedPolymorphismTest {
+    private val objectMapper = jsonMapper()
+    private val writer = objectMapper.writerWithDefaultPrettyPrinter()
+
+    @Test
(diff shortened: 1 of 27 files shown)


zettelmj:
@cjbooms Adding some end2end tests for the generated models. I think while not the full coverage there is some decent coverage aged by this change (increasing the size of the change will most likely only just reduce revieability)

end2end-tests/models-jackson/src/test/resources/deep_nested_polymorphic/third_level_child11.json
```diff
@@ -0,0 +1,16 @@
...
+  "metadata" : {
+    "obj" : {
+      "filed1" : "field 1",
+      "field2" : "field 2"
+    }
+  },
+  "creationDate" : 1,
+  "secondLevelDiscriminator" : "thirdLevelChild1",
+  "rootDiscriminator" : "firstLevelChild",
+  "firstLevelDiscriminator" : "secondLevelChild1"
```
cjbooms, line 15: Why are  3 different levels of discriminator all at the root level in this json?

end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
```diff
@@ -0,0 +1,57 @@
+package com.cjbooms.fabrikt.models.jackson
+
+import com.example.models.*
+import com.fasterxml.jackson.module.kotlin.jsonMapper
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+class DeepNestedPolymorphismTest {
```
cjbooms, line 8: This does not appear to test deep json trees containing nested polymorphic structures. It seems to test individual branches in isolation

cjbooms:
Nice work, it'll be good to have some validated json examples that correspond to the openApi3 specs.

end2end-tests/models-jackson/src/test/resources/deep_nested_polymorphic/third_level_child11.json
zettelmj: That is one to one what is generated during the serialization. In case there isn't something wrong with the setup that is.

end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
zettelmj: I think the deep nesting is on the isomorphism side not the object tree side.

cjbooms: Gotcha, reading the spec I understand that now.

end2end-tests/models-jackson/src/test/resources/enums/enum_ref.json
```diff
@@ -0,0 +1,8 @@
+{
+  "array_of_enums" : null,
+  "inlined_enum" : null,
+  "inlined_extensible_enum" : null,
+  "enum_ref" : "one",
+  "extensible_enum_ref" : null,
+  "list_enums" : null
```
cjbooms, line 7: do you need all the nulls in these examples? Might be cleaner to just leave them absent

end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DeepNestedPolymorphismTest.kt
cjbooms: Another plus for having actual json examples :)

cjbooms (approved):
LGTM - Nit around nulls, but feel free to merge as is

PR #302, opened 2024-07-19
Add an option to add spring ResponseEntity for feign client return type

y-marion:
This adds a configuration option for "http-client-opts" to configurably add the Spring ResponseEntity<> generic around the response type of feign client functions.

This enables the option to get the status code and headers from the response of the external server.

This currently imports  the typeName for ResponseEntity from "com.cjbooms.fabrikt.generators.controller.metadata.SpringImports.RESPONSE_ENTITY" as it was previously only needed in the controllers and i didn't want to duplicate it.
Maybe a different location would be better?

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ This section documents the available CLI parameters for controlling what gets ge
 |                               | CHOOSE ANY OF: |
 |                               |   `RESILIENCE4J` - Generates a fault tolerance service for the client using the following library "io.github.resilience4j:resilience4j-all:+" (only for OkHttp clients) |
 |                               |   `SUSPEND_MODIFIER` - This option adds the suspend modifier to the generated client functions (only for OpenFeign clients) |
+|                               |   `SPRING_RESPONSE_ENTITY_WRAPPER` - This option adds the Spring-ResponseEntity generic around the response to be able to get response headers and status (only for OpenFeign clients). |
 |   `--http-client-target`      | Optionally select the target client that you want to be generated. Defaults to OK_HTTP |
 |                               | CHOOSE ONE OF: |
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -188,60 +188,61 @@ data class Responses(
 
 This section documents the available CLI parameters for controlling what gets generated. This documentation is generated using: `./gradlew printCodeGenUsage`
```
cjbooms, line 190: See line above, please regenerate this from the printCodeGenUsage command. It's not as pretty a table, but keepa code and instructions somewhat in sync

cjbooms:
LGTM.. I'd say that import is fine where it is

README.md
y-marion: sorry about that, missed that instruction there. i've fixed it.

cjbooms (approved):

cjbooms:
This is failing to publish to Sonatype OSSRH.
```
   > Could not PUT 'https://s01.oss.sonatype.org/service/local/staging/deploy/maven2/com/cjbooms/fabrikt/16.6.0/fabrikt-16.6.0.jar'. Received status code 400 from server: Bad Request
```

I'll wait a few days and try again. They seem to be chahing some things lately, as the last time I also had issues and had to modify some stuff. I will wait a few days and try again, in the hope someone else sorts it out. But if this release is urgent for you, respond here and I will dig a bit deeper and create a support request

y-marion:
The release is not that urgent for us, as we do have a temporary fix in place (which is not that pretty, also the reason why i proposed this change) - i just saw that it worked now?

Thank you for your trust and I'll try to help where i can :)

PR #304, opened 2024-07-27
Adds an option to generate spring controllers that return CompletableFutures

zettelmj:
Adds an option to generate spring controllers that return CompletableFutures


src/main/kotlin/com/cjbooms/fabrikt/generators/controller/metadata/Spring.kt
```diff
@@ -41,6 +43,8 @@ object SpringImports {
 
     val AUTHENTICATION = ClassName(Packages.SPRING_AUTHENTICATION, "Authentication")
 
+    val COMPLETABLE_FUTURE = ClassName(Packages.JAVA_UTIL_CONCURRENT, "CompletableFuture")
```
cjbooms, line 46: Would the interface be better? eg CompletionStage

zettelmj: Probably cleaner, although there is nothing else in the jdk the extends the interface.

zettelmj: changed

cjbooms:
👍

cjbooms:
LGTM, just need to regenerate and update the able in the README

zettelmj:
> LGTM, just need to regenerate and update the able in the README. When do you need it released, is it urgent?

I don't need it, I just got asked how to configure fabrikt so that CompletableFutures are returned. Since I had the time it felt like a useful addition.

cjbooms (approved):

PR #305, opened 2024-07-31
Add new `x-async-support` extension for generating async controllers

jeet23:
Recently a new capability to generate `CompletionStage` wrapped controllers was added in #304. However, it was using an `--http-controller-opts` option which does it for all or none controllers.

There are situations where clients require CompletionStage for only specific endpoints in a given controller. This PR adds a fine-grained control to specify which endpoint requires CompletionStage.

OpenAPI's [extensions](https://swagger.io/docs/specification/openapi-extensions/) are used to denote it. If you decorate the path of an endpoint with `x-async-support: true`, the controller generated will be wrapped with `CompletionStage`.


src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerInterfaceGenerator.kt
```diff
@@ -78,7 +79,8 @@ class SpringControllerInterfaceGenerator(
             .addSpringFunAnnotation(op, verb, path.pathString)
             .addSuspendModifier()
 
-        val funcSpec = if (options.contains(ControllerCodeGenOptionType.COMPLETION_STAGE)) {
+        val asyncSupport = op.extensions.containsKey(EXTENSION_ASYNC_SUPPORT)
+        val funcSpec = if (options.contains(ControllerCodeGenOptionType.COMPLETION_STAGE) || asyncSupport) {
```
cjbooms, line 83: Should we allow asyncSupport to override default?
`x-async-support: false`

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerInterfaceGenerator.kt
```diff
@@ -78,7 +79,9 @@ class SpringControllerInterfaceGenerator(
             .addSpringFunAnnotation(op, verb, path.pathString)
             .addSuspendModifier()
 
-        val funcSpec = if (options.contains(ControllerCodeGenOptionType.COMPLETION_STAGE)) {
+        val asyncSupport =
```
cjbooms, line 82: Would rather see this changed to something like below:
```
val explicitAsyncSupport = op.extensions[EXTENSION_ASYNC_SUPPORT] as? Boolean
val asyncSupport = explicitAsyncSupport ?: options.contains(ControllerCodeGenOptionType.COMPLETION_STAGE)

val funcSpec = if (asyncSupport) {...

```

jeet23: Done

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/SpringControllerInterfaceGenerator.kt
jeet23: Added the support for handling both true and false.

PR #307, opened 2024-08-05
Fix BigDecimal defaults.

### LINKED ISSUE #306: Default Value for BigDecimal is generated as an Integer
The default value for a number type (represented as a BigDecimal should use the constructor when setting default value

Error:
` The integer literal does not conform to the expected type BigDecimal.`

Generated:
Screenshot 2024-08-04 at 9 11 41 PM

schema:
```
markupDecimal:
    default: 0
    type: number
    multipleOf: 0.01
```

cjbooms:
Closes #306


cjbooms:
Quick and dirty fix, if anyone wants to run their eyes over it

PR #314, opened 2024-09-13
Add client streaming flag protection

averabaq:
It only allows to override `byte-array` to `input-stream` data type generation on server-side code. This override is ignored on the client side until full streaming is supported.

```
WARNING: Client code generation does not support streaming, yet. The override flag 
'BYTEARRAY_AS_INPUTSTREAM' is ignored. If generating server side code, please consider 
splitting the client & server in different fabrikt executions. Defaulting to `ByteArray`...
```


cjbooms (approved):
LGTM

PR #321, opened 2024-10-06
Allow having multiple authentication alternatives in Ktor controllers

lunakoly:
`authenticate()` accepts a vararg and by default allows the API access if at least one option is satisfied.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
@@ -173,17 +173,18 @@ class KtorControllerInterfaceGenerator(
 
         val addAuth = securityOption.allowsAuthenticated && options.contains(ControllerCodeGenOptionType.AUTHENTICATION)
         if (addAuth) {
-            // Using the key from the first security requirement as the auth name
-            val authName = if (operation.hasSecurityRequirements()) {
-                operation.securityRequirements.first().requirements.keys.first()
+            val authNames = if (operation.hasSecurityRequirements()) {
+                operation.securityRequirements
+                    .filter { it.requirements.isNotEmpty() }
+                    .joinToString(", ") { "\"" + it.requirements.keys.first() + "\"" }
             } else {
                 // Fall back to the global security requirements
-                this.api.openApi3.securityRequirements.first().requirements.keys.first()
+                "\"" + this.api.openApi3.securityRequirements.first().requirements.keys.first() + "\""
             }
 
             builder
                 .addStatement(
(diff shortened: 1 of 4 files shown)


lunakoly:
This should be allowed, because the spec permits it ([§ Security Requirement Object](https://swagger.io/specification/)).

> When a list of Security Requirement Objects is defined on the [OpenAPI Object](https://swagger.io/specification/#openapi-object) or [Operation Object](https://swagger.io/specification/#operation-object), only one of the Security Requirement Objects in the list needs to be satisfied to authorize the request.

Additional clarifications: [issues 4119](https://github.com/OAI/OpenAPI-Specification/issues/4119), [§ Using Multiple Authentication Types](https://swagger.io/docs/specification/v3_0/authentication/#using-multiple-authentication-types) from OpenAPI Guide

cjbooms:
@ulrikandersen do you have any opinion on this as the original contributor of the KTOR controllers? 
Looks fine to me, but I have never used Ktor.

ulrikandersen:
It also looks good to me!

It would be nice to also support [other combinations](https://swagger.io/docs/specification/v3_0/authentication/#using-multiple-authentication-types), such as `BasicAuth OR BearerAuth`, but that can be done in separate future PRs.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
```diff
@@ -173,17 +173,18 @@ class KtorControllerInterfaceGenerator(
 
         val addAuth = securityOption.allowsAuthenticated && options.contains(ControllerCodeGenOptionType.AUTHENTICATION)
         if (addAuth) {
-            // Using the key from the first security requirement as the auth name
-            val authName = if (operation.hasSecurityRequirements()) {
-                operation.securityRequirements.first().requirements.keys.first()
+            val authNames = if (operation.hasSecurityRequirements()) {
```
ulrikandersen, line 176: I think this reads slightly better, but it is minor detail:

```kotlin
val authNames = if (operation.hasSecurityRequirements()) {
    operation.securityRequirements
        .filter { it.requirements.isNotEmpty() }
        .map { it.requirements.keys.first() }
} else {
    // Fall back to the global security requirements
    listOf(this.api.openApi3.securityRequirements.first().requirements.keys.first())
}.joinToString(", ") { "\"$it\"" }
```

ulrikandersen (approved):

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
cjbooms: if you switch it to a suggestion, I can commit it

PR #324, opened 2024-10-09
Fix under-/overflow in min/max validation for controller parameters

y-marion:
The javax/jakarta annotation for validating input parameters in header, query or path of a controller were wrongly truncated to an Integer, even though a long is allowed, and we were correctly generating the type parameter as Long.

Javax/Jakarta both specify that the value of the `@Min` / `@Max` annotation is of type `long` (and we were also adding the member as `%L` anyway), so this only removes the `toInt` call and replaces it with a `toLong` call.

I've added a testcase for this; The problem can be seen if only the first commit is checked out and the tests are then executed.

Additionally, we'd need to think about a more general fix for this, as this doesn't allow decimal value minimum/maximum in the params (they are just truncated to Integer values at the moment); either we drop the Min/Max Annotation completely in favor of DecimalMin/DecimalMax (which also allow the `exclusive`  restriction then), or decide based on the type of value in the min/max validation.


cjbooms:
Deciding based on type whether to use Min Max or DecimalMin DecimalMax seems like the best solution. 

But obviously truncating decimals to integers is bad, so we should at least stop that from happening, possibly just by omitting the annotations from decimals if that is the easier path .

Happy to merge this improvement first, but can we move the test cases in under `test/resources/examples/jakartaValidationAnnotations` rather than having a new folder.

y-marion:
Moved the testcases. 
I'll look into fixing the truncation for the decimals next, but not sure when i'll have time for it; probably in the coming 1-2 weeks; i don't think it would be that difficult to decide based on the type which annotation we'll need.

cjbooms:
> Moved the testcase. I'll look into fixing the truncation for the decimals next, but not sure when i'll have time for it; probably in the coming 1-2 weeks; i don't think it would be that difficult to decide based on the type which annotation we'll need.

Nice one. Do you need a release cut for this today, or will we delay until the better fix is ready?

y-marion:
It would be convenient if we could cut a release for this, as i'm not exactly sure when i'll get to the better fix. We don't have a usecase for decimal parameters (yet), but we do have a current issue with the over/underflow 😅.

PR #334, opened 2024-11-25
Add usage tips to README

ulrikandersen:
This PR aims to enhance the Fabrikt README by adding helpful pointers to ensure users can get the most out of the tool 🚀.

There are a few well-known gotchas that have been raised in issues, so we thought it would be valuable to document them here. This should reduce the chances of users giving up on Fabrikt with the impression that “It’s not working!” and instead help them troubleshoot effectively.

Any other gotchas worth mentioning? 

Cc/ @cjbooms, @acanda, @averabaq, ..?

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,12 +170,9 @@ The [exec-maven-plugin](http://www.mojohaus.org/exec-maven-plugin/examples/examp
 ## Getting the Most from Fabrikt
 
 ### 1. Prefer components to inline schemas
-
 While inline schemas are perfectly valid they are not supported by Fabrikt in all circumstances.
 This is especially true for request bodies and non-trivial parameters. Instead, define your schemas in the
-components section of the OpenAPI spec (`components.parameters` & `components.requestBodies`).
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,6 +167,20 @@ fabrikt {
 
 The [exec-maven-plugin](http://www.mojohaus.org/exec-maven-plugin/examples/example-exec-using-plugin-dependencies.html) is capable of downloading the Fabrikt library from Maven Central and executing its main method with defined arguments.
 
+## Getting the Most from Fabrikt
+
+### 1. Prefer components to inline schemas
+
+While inline schemas are perfectly valid they are not supported by Fabrikt in all circumstances.
+This is especially true for request bodies and non-trivial parameters. Instead, define your schemas in the
+components section of the OpenAPI spec (`components.parameters` & `components.requestBodies`).
```
cjbooms, line 176: Uneeded line break

cjbooms (approved):
LGTM. I immediately looked to see if you had covered anything about inline generation. So delighted to see these additions

PR #336, opened 2024-11-26
Upgrade to Ktor Server 3.0.1

ulrikandersen:
This PR adapts the Ktor Controller Interface Generator to support Ktor Server 3.0.1.

Upgrading to Ktor Server 3.0.1 means letting go of the 'principal' parameter to the controller functions. The [Principal interface](https://api.ktor.io/ktor-server/ktor-server-plugins/ktor-server-auth/io.ktor.server.auth/-principal/index.html) is no longer available and thus we cannot provide a common supertype of the different Principals. Instead the user should get the principal with `call.principal<P>()` in the implementation. This will work just fine with Ktor 2.3.x too.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -40,8 +40,8 @@ allprojects {
 
 val jacksonVersion by extra { "2.15.1" }
 val junitVersion by extra { "5.9.2" }
-val ktorVersion by extra { "2.3.9" }
 val kotlinxSerializationVersion by extra { "1.6.3" }
+val ktorVersion by extra { "3.0.1" }
 val kotlinxDateTimeVersion by extra { "0.6.1" }
 
 dependencies {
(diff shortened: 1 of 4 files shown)


cjbooms (approved):
Apologies for the delay on this Ulrik. Feel free to merge if you are happy, I am not an expert on Ktor. Just ping me if you want a new major release cut, I can have on ready in 30 minutes or so

ulrikandersen:
> Apologies for the delay on this Ulrik. Feel free to merge if you are happy, I am not an expert on Ktor. Just ping me if you want a new major release cut, I can have on ready in 30 minutes or so

No worries at all ✌️ I am not in a hurry to get a new major version out, but I will ping in case that changes. I am also working on a fix for `kotlinx.serialization` that could potentially be included in an upcoming major version release.

PR #337, opened 2024-11-26
Upgrade to Kotlin 2.0.20

ulrikandersen:
Upgrades Kotlin to 2.0.20. Shouldn't change anything, but we might as well keep up with the latest major version of Kotlin.

`Kotlinx.Serialization` is also upgraded as `1.7.3` [uses 2.0.20 by default](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.7.3). This affects only the end2end test.


cjbooms:
Nice one. We should try and do Java and Gradle soon too. It's been awhile for both

cjbooms (approved):

PR #338, opened 2024-12-10
Implement support for optional objects in openapi 3.1.0

nils-degroot:
Currently mapping from openapi 3.1.0 nullable types only works with the types array which does not support object. This pr provides mapping for `oneOf (object | null)` to the legacy style mapping. See the following input openapi spec:

```yaml
openapi: 3.1.0
components:
  schemas:
    NullableObject:
      type: object
      properties:
        field:
          oneOf:
            - type: 'null'
            - $ref: '#/components/schemas/TheObject'
    TheObject:
      properties:
        name:
          type: string
```

It would be updated to this openapi spec after this pr:

```yaml
openapi: 3.1.0
components:
  schemas:
    NullableObject:
      type: object
      properties:
        field:
          allOf:
            - $ref: '#/components/schemas/TheObject'
          nullable: true
    TheObject:
      properties:
        name:
          type: string
```

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/YamlUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/YamlUtils.kt
@@ -77,6 +77,13 @@ object YamlUtils {
                         (node as ObjectNode).replace("type", nonNullType)
                         requiresNullable = true
                     }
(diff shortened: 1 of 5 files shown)


cjbooms:
Nice one. Thanks for the contribution

PR #340, opened 2024-12-17
Replace values() with entries() with in enums

Palleas:
Hi there - I was digging into the generated code and I realized there was a recommendation to replace `values()` with `entries` when using enums: https://www.jetbrains.com/help/inspectopedia/EnumValuesSoftDeprecate.html

Since I wanted to contribute back to this library that we started using recently, I figured it was worth a shot. The tests are failing right now because I couldn't figure out how to rebuild the expected models in `src/test/resources/examples`, if you can point me toward the way to do that, I'll finish the work on this pull-request, assuming that's something you'd like.


cjbooms:
I'm afraid it's a manual task, we never built anything that overwrites them with the new value. Use find and replace, or you could try to add a function to the tests to correct the  files

Palleas:
Oh cool, at least that means I haven't missed anything! I'll do a manual replace first and maybe I can contribute something in a future PR, if that works for you?

Palleas:
Updated!

cjbooms (approved):

PR #342, opened 2024-12-20
Omit discriminator property for Kotlin serialization

ulrikandersen:
Discriminated oneOf models generated with Kotlin serialization will currently not work because the discriminator property causes a name conflict with Kotlin Serialization's class discriminator:

`Sealed class 'subclass' cannot be serialized as base class 'com.example.models.Superclass' because it has property name that conflicts with JSON class discriminator 'type'.`

The issue was not discovered earlier because the OpenAPI spec for end-2-end test for Kotlin serialization models did not define the discriminator property on the child schemas. This PR adds the missing `type` property to the test.

In order for the generated code to be valid with Kotlin serialization we must omit the property in the models. The solution in this PR is to simply skip adding the property as part of the code generation process.

Skipping properties in the generation process has the side effect that if the discriminator is the only property on the schema we are left with a class with zero properties which must then converted to an object ([example](src/test/resources/examples/discriminatedOneOf/models/kotlinx/StateA.kt)). It would be more clean to filter out the property before the generation step, but I am not sure where we would hook in in order to do so.

I am really open to feedback on this approach and to any suggestions for improvements.

### DIFF
--- a/end2end-tests/models-kotlinx/openapi/api.yaml
+++ b/end2end-tests/models-kotlinx/openapi/api.yaml
@@ -71,9 +71,12 @@ components:
     LandlinePhone:
       type: object
       required:
+        - type
         - number
         - area_code
       properties:
+        type:
+          type: string
         number:
           type: string
         area_code:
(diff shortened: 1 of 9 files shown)


cjbooms (approved):
LGTM

PR #349, opened 2025-01-14
Update README.md

cjbooms:
Let's see if including a screenshot of the UI looks good


cjbooms:
@ulrikandersen what do you think about adding a clickable screenshot?

ulrikandersen (approved):
That’s a great idea! Makes it stand out more. Let’s do it.

PR #359, opened 2025-02-02
Add option to add spring-cloud-openfeign-starter annotation

### LINKED ISSUE #358: Add option to add spring-cloud-starter-openfeign @FeignClient annotation
spring-cloud-starter-openfeign has an annotation `@FeignClient`,
https://javadoc.io/doc/org.springframework.cloud/spring-cloud-openfeign-core/3.1.9/org/springframework/cloud/openfeign/FeignClient.html

Can we add an option to add this annotation when generating openfeign client?

mirror-kt:
closes #358


src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
```diff
@@ -53,6 +53,16 @@ class OpenFeignInterfaceGenerator(
 
             val clientType = TypeSpec.interfaceBuilder(simpleClientName(resourceName))
                 .addAnnotation(AnnotationSpec.builder(Suppress::class).addMember("%S", "unused").build())
+                .apply {
+                    if (options.contains(ClientCodeGenOptionType.SPRING_CLOUD_OPENFEIGN_STARTER_ANNOTATION)) {
+                        addAnnotation(
+                            OpenFeignAnnotations.feignClientBuilder()
+                                .addMember("configuration = %S", api.openApi3.info.getExtension("x-feign-client-name"))
```
mirror-kt, line 60: I've tried using vender options (x- prefixed), but any other suggestions?

cjbooms: I think this approach is good. We also use it for Jackson annotations. But what happens is the options is turned on, but the extension is not present?

cjbooms:
Hi, sorry for the delay. Open to accepting this contribution but I do not know anything about Feign.  Can you please flesh out the PR description a bit more to give some hints.
For instance:
 - What does this annotation do, and why is it important?
 - What parameters for the annotation are required and what are optional? What should fabrikt support and should they be all individual x- extentions?
 - Will there be a case where for a single API spec, multiple independent clients are generated, and some require the annotation and others do not?

mirror-kt:
Thanks for taking a look.

> What does this annotation do, and why is it important?
This is not OpenFeign itself, but an annotation related to spring-cloud-starter-openfeign, a library for integration with spring-boot.

OpenFeign clients with this annotation are automatically bean registered by SpringBoot (just like `@Service` and `@Component`).

See also: https://docs.spring.io/spring-cloud-openfeign/docs/current/reference/html/#netflix-feign-starter

> What parameters for the annotation are required and what are optional? What should fabrikt support and should they be all individual x- extentions?

We need `name` and `contextId`.

I was thinking of getting the name from x- extension and using resourceName for contextId. contextId could also be obtained from extension.

I just looked at it and found a mistake in the code, so the merge will have to wait.

> Will there be a case where for a single API spec, multiple independent clients are generated, and some require the annotation and others do not?

I do not believe it is necessary.
There may be the ability to filter out clients that are generated regardless of annotations, but that should be worked in another PR.

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
mirror-kt: This value should be the same for all Clients generated in a single fabrikt run, and the baseUrl, retryPolicy, etc. should be set based on this value.

It cannot contain whitespace, and lowerCamelCase is ideal.

Any suggestions for a good default value?

settings.gradle.kts
```diff
@@ -1,3 +1,7 @@
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.9.0"
```
cjbooms, line 2: What does this do? Seems unused

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
cjbooms: If it's always the same, why not pass it in via a fabrikt config option, similar to:
` --http-model-suffix`

`x-` extensions make sense when the value needs to change in different parts of the spec, but it sounds like global external config is sufficient for you. 

I'm not sure about sensible defaults, I have never used this.

settings.gradle.kts
mirror-kt: It's a plugin that downloads the necessary JDKs depending on the Gradle Toolchain configuration.
It is not directly related to this change, so you can revert it, but if you plan to accept contribute, I think it is better to include it.

mirror-kt: https://github.com/gradle/foojay-toolchains

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
mirror-kt: Okay, it certainly sounds like a global external config would be better.

I'm on the road right now, so I'll make a commit to fix it when I get back.

mirror-kt: done.
It is now get as a cli option.

settings.gradle.kts
mirror-kt: reverted it

cjbooms:
Nice one. If you want to, you can add the config to one of the existing test examples to see if it works the way you expect. No need to create an entire new test case.

cjbooms (approved):

mirror-kt:
>  If you want to, you can add the config to one of the existing test examples to see if it works the way you expect.

I added it. If it looks OK, please merge it in!

cjbooms:
Moved to #365

PR #360, opened 2025-02-03
Fix correctly defined nullable enums causing a NPE

### LINKED ISSUE #357: Nullable enums cause NPE
Hello,
I was trying to generate the client and model code for the German Brightsky API: https://brightsky.dev/docs/

Their OpenAPI Spec works flawlessly in the Swagger Editor, but when trying to generate the code with fabrikt, it simply fails due to 

```
Caused by: java.lang.NullPointerException: Cannot invoke "Object.toString()" because "it" is null
	at com.cjbooms.fabrikt.util.KaizenParserExtensions.getEnumValues(KaizenParserExtensions.kt:94)
	at com.cjbooms.fabrikt.model.KotlinTypeInfo$Companion.from(KotlinTypeInfo.kt:96)
	at com.cjbooms.fabrikt.model.PropertyInfo$Field.<init>(PropertyInfo.kt:170)
	at com.cjbooms.fabrikt.model.PropertyInfo$Companion.getInLinedProperties(PropertyInfo.kt:127)
	at com.cjbooms.fabrikt.model.PropertyInfo$Companion.topLevelProperties(PropertyInfo.kt:57)
	at com.cjbooms.fabrikt.generators.model.ModelGenerator.createModels(ModelGenerator.kt:179)
	at com.cjbooms.fabrikt.generators.model.ModelGenerator.generate(ModelGenerator.kt:162)
	at com.cjbooms.fabrikt.cli.CodeGenerator.models(CodeGenerator.kt:65)
	at com.cjbooms.fabrikt.cli.CodeGenerator.generateClient(CodeGenerator.kt:55)
	at com.cjbooms.fabrikt.cli.CodeGenerator.generateCode(CodeGenerator.kt:36)
	at com.cjbooms.fabrikt.cli.CodeGenerator.generate(CodeGenerator.kt:32)
	at com.cjbooms.fabrikt.cli.CodeGen.generate(CodeGen.kt:84)
	at com.cjbooms.fabrikt.cli.CodeGen.main(CodeGen.kt:22)
	at ch.acanda.gradle.fabrikt.generator.FabriktGeneratorKt.generate(FabriktGenerator.kt:18)
	... 119 more
```

When I removed all `null` values from all the enums in the YAML it worked.

A quick debugging session showed that the issue is there: https://github.com/cjbooms/fabrikt/blob/d538e020cf12f7eaacd09bf1c117363c77cb8afb/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt#L94

So that `toString()` call there needs to use the correct extension function for `Any?`, and not the member method of `Any`. I don't know if that is enough, though.

BUT: Even after removing all nullable enums, the generated client code (for OkHTTP) did not compile still. It tried to provide a `List<Int>` to a function that only accepts `List<String>`. That is the case for all the `source_id` parameters in the API spec.

PS: I was using a Gradle build with the latest version 1.11.0 of the plugin, which in turn uses fabrikt 20.0.0.

xehmer:
Closes #357

It's simply fixing the NPE that causes the whole generation to fail if an enum has `null` in its list of allowed values, which is how nullable enums should be modeled if done correctly.

Nullable is still in a strange place in OpenAPI per se, since it's completely separate from `required` on the containing object property level. You can actually define some property as `required` even if that property is nullable per it's schema. That would mean that the property key in the JSON would need to be there, but the value could be null. And that is different from the property being completely left out as it would be possible if the property was not `required`. 

The main issue there is that that would be quite hard to model correctly in Kotlin. You would actually need some kind of `NULL` value within an enum that is nullable, to be able to seperate the two cases of it being null and the property being left our wholly. In the containing object, the reference to that enum being Kotlin-nullable would mean "property is optional", and that artificial `NULL` enum value would mean "property value should be null, but the property key should be there".

So with my change it's just working "correctly" for nullable enums that are optional as well, which is most likely the only real-world scenario that should occur anyway.

### DIFF
--- a/src/test/resources/examples/enumExamples/api.yaml
+++ b/src/test/resources/examples/enumExamples/api.yaml
@@ -33,6 +33,8 @@ components:
           $ref: '#/components/schemas/ExtensibleEnumObject'
         list_enums:
           $ref: '#/components/schemas/ListEnums'
+        nullable_ref:
+          $ref: '#/components/schemas/NullableEnum'
 
     EnumObject:
       type: string
@@ -85,5 +87,7 @@ components:
           bar_prop:
             type: string
 
-
-
+    NullableEnum:
+      type: string
+      nullable: true
+      enum: [ "foo", "bar", null ]
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
@@ -91,7 +91,7 @@ object KaizenParserExtensions {
 
     @Suppress("UNCHECKED_CAST")
     fun Schema.getEnumValues(): List<String> = when {
-        this.hasEnums() -> this.enums.map { it.toString() }.filterNot { it.isBlank() }
+        this.hasEnums() -> this.enums.filterNotNull().map { it.toString() }.filterNot { it.isBlank() }
         !MutableSettings.modelOptions().contains(ModelCodeGenOptionType.X_EXTENSIBLE_ENUMS) -> emptyList()
         else -> extensions[EXTENSIBLE_ENUM_KEY]?.let { it as List<String?> }?.filterNotNull()
             ?.filterNot { it.isBlank() } ?: emptyList()
--- a/src/test/resources/examples/enumExamples/models/EnumHolder.kt
+++ b/src/test/resources/examples/enumExamples/models/EnumHolder.kt
@@ -22,4 +22,7 @@ public data class EnumHolder(
   @param:JsonProperty("list_enums")
   @get:JsonProperty("list_enums")
   public val listEnums: List<ContentType>? = null,
+  @param:JsonProperty("nullable_ref")
+  @get:JsonProperty("nullable_ref")
+  public val nullableRef: NullableEnum? = null,
 )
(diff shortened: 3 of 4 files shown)


cjbooms:
Thanks for the contribution. 

This solution fits with the simple approach taken by fabrikt, which is to treat required properties that are nullable as optional, as can be seen in the [optionalVsRequired](https://github.com/cjbooms/fabrikt/blob/d538e020cf12f7eaacd09bf1c117363c77cb8afb/src/test/resources/examples/optionalVsRequired/api.yaml) test case.

Good write-up of the state of null and required. For Jackson, there is some support for required nullable fields, which was used for the merge patch contribution, in the form of:
`org.openapitools.jackson.nullable.JsonNullable`
`JsonNullable<NestedMergePatchInlineInner> = JsonNullable.undefined(),`
We just don't use if for the general case.

src/test/resources/examples/enumExamples/api.yaml
```diff
@@ -33,6 +33,8 @@ components:
           $ref: '#/components/schemas/ExtensibleEnumObject'
         list_enums:
           $ref: '#/components/schemas/ListEnums'
+        nullable_ref:
```
cjbooms, line 36: It would be good to see another identical property, but added to the required list

Issue #362, opened 2025-02-04
Add UUID to kotlinx.serialization


alex-rieger:
Hello,

When using kotlinx.serialization strings with `format: uuid` are converted to `String` in models, while Jackson serialization generates `UUID`. I think it would be a good addition if kotlinx.serialization is also able to produce `UUID`. Directly having `UUID` would eliminate the need to convert `UUID <> String` when working with e.g. Exposed.

Is this something that could be done? I'd be happy to open a pull request for it.

I found the following code in `kotlin/com/cjbooms/fabrikt/model/KotlinTypeInfo.kt`:
```kotlin
 OasType.Uuid -> {
  if (MutableSettings.serializationLibrary() == KOTLINX_SERIALIZATION) Text // could possibly be Kotlin native UUID once that becomes stable
  else Uuid
}
```
As far as I'm aware Kotlin's Uuid is not stable yet. Maybe this feature could be added as Type Override so users can choose whether it will be UUID / String (Uuid in future)? 

Examples
```yaml
openapi: 3.1.0

components:
  schemas:
    Pet:
      type: object
      required:
        - id
      properties:
        id:
          type: string
          format: uuid
```

```kotlin
// kotlinx.serialization
package com.example.models

import javax.validation.constraints.NotNull
import kotlin.String
import kotlinx.serialization.SerialName
import kotlinx.serialization.Serializable

@Serializable
public data class Pet(
  @SerialName("id")
  @get:NotNull
  public val id: String,
)

// Jackson serialization
package com.example.models

import com.fasterxml.jackson.`annotation`.JsonProperty
import java.util.UUID
import javax.validation.constraints.NotNull

public data class Pet(
  @param:JsonProperty("id")
  @get:JsonProperty("id")
  @get:NotNull
  public val id: UUID,
)
```

ulrikandersen:
Hi @alex-rieger,

Since kotlinx.serialization does not have a built in serializer for UUID it was decided to ignore the `format` and fall back to string. The same applies for `format: uri` (and actually also for `format: number` which currently maps to `BigDecimal`, which will not compile and needs to be fixed).

```
Serializer has not been found for type 'UUID'. 
To use context serializer as fallback, explicitly annotate type or property with @Contextual
```

It is possible to provide a custom serializer for these types using the `Contextual` ([docs](https://kotlinlang.org/api/kotlinx.serialization/kotlinx-serialization-core/kotlinx.serialization/-contextual/)) annotation. 

[...]

PR #364, opened 2025-02-06
Kotlin serialization: Numbers

ulrikandersen:
With Jackson we use Java's BigDecimal for numbers without a format keyword, but because Java types are not serializable with Kotlin Serialization we need to use a Kotlin native type instead. ~~I believe the best we can do is Double although it has lower precision than BigDecimal.~~

Alternatively we could allow the user to provide his own custom serializer for BigDecimal (and other Java types like UUID and URI) as [discussed in this issue](https://github.com/cjbooms/fabrikt/issues/362) by annotating the field with `@Contexual`.

**Update:** 

In order to not loose precision we annotate the field with `@Contexual` and leave it to the user to provide a custom serializer. This is aligns with the [recommended approach](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#serializing-large-decimal-numbers) in the documentation. 

We can add the same possibility for string formats (`uuid`, `uri`, `base64`, `binary`), but because no precision is lost we could add it as a configuration option. I prefer to do that in a future PR.

### DIFF
--- a/end2end-tests/models-kotlinx/build.gradle.kts
+++ b/end2end-tests/models-kotlinx/build.gradle.kts
@@ -1,7 +1,6 @@
 val fabrikt: Configuration by configurations.creating
 
 val generationDir = "$buildDir/generated"
-val apiFile = "$projectDir/openapi/api.yaml"
 
 sourceSets {
     main { java.srcDirs("$generationDir/src/main/kotlin") }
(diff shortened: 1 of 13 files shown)


cjbooms:
I vote for @Contextual and letting the users define their own serializer

Not a fan of the pending blog post about a fabrikt being the root cause of a some international banking incident due to loss of precision.

ulrikandersen:
> I vote for @contextual and letting the users define their own serializer
> 
> Not a fan of the pending blog post about a fabrikt being the root cause of a some international banking incident due to loss of precision.

Same here 😬 I'd rather see some positive blog posts 😌

I've updated the PR to add `@Contextual` and force the user to provide a serializer.

cjbooms:
LGTM - Sorry for the delay

cjbooms (approved):

PR #367, opened 2025-02-13
Kotlin serialization: Use String for byte and binary

ulrikandersen:
The **primary change** in this PR is that strings with format `byte` and `binary` are treated as Strings with Kotlin serialization. This aligns with the current behaviour for UUID and URI.

```kotlin
OasType.Base64String -> {
    if (MutableSettings.serializationLibrary() == KOTLINX_SERIALIZATION) Text // no native ByteArray in kotlinx.serialization
    else ByteArray
}
OasType.Binary -> {
    if (MutableSettings.serializationLibrary() == KOTLINX_SERIALIZATION) Text // no native Binary in kotlinx.serialization
    else getOverridableByteArray()
}
```

The current implementation where `byte` and `binary` translated to `ByteArray` does not work because `ByteArray` cannot be serialized without a [custom serializer](https://github.com/Kotlin/kotlinx.serialization/blob/master/docs/json.md#base64).

Please note: This change is effectively cancelled out by #368 if that gets merged, but could still be seen as a step in the direction of making the generation accurate.

**Secondary change:** To ensure we test primitive types for both Jackson and Kotlin Serialization an end2end test for Jackson is added.


ulrikandersen:
@cjbooms I’m closing this in favor of #368, which overrides the main change in this PR.

PR #368, opened 2025-02-13
String format type overrides

ulrikandersen:
To be merged after #367.

**Breaking Change:** The default behavior when using Kotlin Serialization is changed. Previously, the `format` on string types was ignored because custom serializers were required. With this update, Fabrikt will now generate the corresponding Java types by default, requiring users to register custom serializers. This change fully leverages the JVM type system.  

To accommodate non-JVM use cases and scenarios where custom serializers are not desired, an escape hatch has been introduced via a new set of type overrides:  

- `BYTE_AS_STRING` – Treats `byte` format as a `String`, ignoring the format  
- `BINARY_AS_STRING` – Treats `binary` format as a `String`, ignoring the format  
- `URI_AS_STRING` – Treats `uri` format as a `String`, ignoring the format  
- `UUID_AS_STRING` – Treats `uuid` format as a `String`, ignoring the format  
- `DATE_AS_STRING` – Treats `date` format as a `String`, ignoring the format  
- `DATETIME_AS_STRING` – Treats `date-time` format as a `String`, ignoring the format  

When these overrides are enabled, the `format` is discarded, and `String` is used for the field instead. This applies to both Kotlin Serialization and Jackson.

In order for the end2end tests to pass a set of [custom serializers](https://github.com/cjbooms/fabrikt/tree/bc1cdcb4dc10a65678a274aa882079fccb30b0f5/end2end-tests/models-kotlinx/src/test/kotlin/com/cjbooms/fabrikt/models/kotlinx/serializers) are present in the test directory. These will probably be useful for end users too, but I don’t know if I am ready to ship them as part of the generated code, but maybe we should?


cjbooms:
Should we also add: 
`INT64_AS_String`

ulrikandersen:
> Should we also add: `INT64_AS_String`

What would the motivation be? Are you trying to solve the need for a custom serializer for `BigDecimal` in Kotlin serialization (that is needed for `number` without a format)?

cjbooms:
> > Should we also add: `INT64_AS_String`
> 
> What would the motivation be? Are you trying to solve the need for a custom serializer for `BigDecimal` in Kotlin serialization (that is needed for `number` without a format)?

Yes, same justification as the String ones. If users want to avoid contextual, could we use string for int64 to avoid loss of precision?

cjbooms:
> > > Should we also add: `INT64_AS_String`
> > 
> > 
> > What would the motivation be? Are you trying to solve the need for a custom serializer for `BigDecimal` in Kotlin serialization (that is needed for `number` without a format)?
> 
> Yes, same justification as the String ones. If users want to avoid contextual, could we use string for int64 to avoid loss of precision?

Actually, I see what you mean now. It's only numbers without a format. Not format int64 .

cjbooms:
> > > > Should we also add: `INT64_AS_String`
> > > 
> > > 


Had a quick look and neither Jackson or Kotlin serialization would even handle this. So forget I said anything :)

ulrikandersen:
> Actually, I see what you mean now. It's only numbers without a format. Not format int64 .

In environments where the mere presence of `import java.math.BigDecimal` would cause compilation errors (Kotlin multi platform) it could be useful to be able ask Fabrikt to treat `number without format` as a String, but that would require a custom serializer/deserializer as it would have to convert from JSON number to string and back. We could do that, but I think we should wait for the need to arise and consider it then.

PR #373, opened 2025-02-17
Handle query params with non primitive type in Ktor controller

### LINKED ISSUE #370: Ktor server - Controller always returns `400` when route has a query parameter of type `DateTime`
Hello! First thanks for the library :)

I'm using the Ktor generator with this OpenApi specification:
```yaml
openapi: 3.0.0
paths:
  /hello:
    get:
      parameters:
        - name: start_date
          in: query
          required: true
          schema:
            type: string
            format: date-time
      responses:
        '200':
          content:
            application/json:
              schema:
                type: object
                properties:
                  message:
                    type: string
                    example: "Hello! The start date is 2023-10-01T00:00:00Z"
```

This generate a controller that handles the `hello` route like this:
```kotlin
public fun Route.helloRoutes(controller: HelloController) {
    `get`("/hello") {
      val startDate =
          call.request.queryParameters.getOrFail<kotlinx.datetime.Instant>("start_date")
      controller.get(startDate, TypedApplicationCall(call))
    }
  }
```

When calling this route, I always get a `400`. Conversion from the query parameter string value to `kotlinx.datetime.Instant` always fails. I think this is due to Ktor's `DefaultConversionService` only supporting primitive types. We can extend this service to handle additional types by using the [DataConversion plugin](https://ktor.io/docs/server-data-conversion.html).

However, in order to make it work, the controller's generated code needs to be updated. Instead of using `DefaultConversionService` directly, we should get an instance of Ktor's `ConversionService` from `call.application.conversionService`.

Before opening a PR, I would like to get your insight regarding what I've just described and the suggested solution.
Thanks in advance 🙏

nsprod:
This PR aims to fix #370 

The fix only applies to optional parameter for now on. I would prefer to fix required query params in another PR if you don't mind.

cc @ulrikandersen

### DIFF
--- a/end2end-tests/ktor/build.gradle.kts
+++ b/end2end-tests/ktor/build.gradle.kts
@@ -69,13 +69,13 @@ tasks {
     val generateKtorInstantDateTimeCode = createCodeGenerationTask(
         "generateKtorInstantDateTimeCode",
         "src/test/resources/examples/instantDateTime/api.yaml",
-        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
+        listOf("--serialization-library", "KOTLINX_SERIALIZATION")
     )
 
     val generateKtorQueryParametersCode = createCodeGenerationTask(
         "generateKtorQueryParametersCode",
         "src/test/resources/examples/queryParameters/api.yaml",
-        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
+        listOf("--serialization-library", "KOTLINX_SERIALIZATION")
     )
 
     withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
--- a/end2end-tests/ktor/build.gradle.kts
+++ b/end2end-tests/ktor/build.gradle.kts
@@ -9,6 +9,7 @@ sourceSets {
 
 plugins {
     kotlin("jvm")
+    kotlin("plugin.serialization") version "2.1.0"
 }
 
 java {
@@ -23,6 +24,7 @@ val ktorVersion: String by rootProject.extra
 dependencies {
     implementation("jakarta.validation:jakarta.validation-api:3.0.2")
     implementation("javax.validation:validation-api:2.0.1.Final")
+    implementation("org.jetbrains.kotlinx:kotlinx-datetime:0.6.2")
     implementation("com.fasterxml.jackson.module:jackson-module-kotlin:$jacksonVersion")
     implementation("com.fasterxml.jackson.core:jackson-databind:$jacksonVersion")
     implementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion")
@@ -32,8 +34,10 @@ dependencies {
     // ktor server
     implementation("io.ktor:ktor-server-content-negotiation-jvm:$ktorVersion")
     implementation("io.ktor:ktor-serialization-jackson:$ktorVersion")
+    implementation("io.ktor:ktor-serialization-kotlinx-json:$ktorVersion")
     implementation("io.ktor:ktor-server-auth:$ktorVersion")
     implementation("io.ktor:ktor-server-status-pages:$ktorVersion")
+    implementation("io.ktor:ktor-server-data-conversion:$ktorVersion")
 
     // ktor test
     testImplementation("io.ktor:ktor-server-test-host:$ktorVersion")
@@ -62,10 +66,17 @@ tasks {
         listOf("--http-controller-opts", "AUTHENTICATION")
     )
 
+    val generateKtorInstantDateTimeCode = createCodeGenerationTask(
+        "generateKtorQueryParametersCode",
+        "src/test/resources/examples/instantDateTime/api.yaml",
+        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
+    )
+
     withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
         kotlinOptions.jvmTarget = "17"
         dependsOn(generateKtorCode)
         dependsOn(generateKtorAuthCode)
+        dependsOn(generateKtorInstantDateTimeCode)
     }
 
 
--- a/end2end-tests/ktor/build.gradle.kts
+++ b/end2end-tests/ktor/build.gradle.kts
@@ -67,8 +67,14 @@ tasks {
     )
 
     val generateKtorInstantDateTimeCode = createCodeGenerationTask(
-        "generateKtorQueryParametersCode",
+        "generateKtorInstantDateTimeCode",
         "src/test/resources/examples/instantDateTime/api.yaml",
+        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
+    )
+
+    val generateKtorQueryParametersCode = createCodeGenerationTask(
+        "generateKtorQueryParametersCode",
+        "src/test/resources/examples/queryParameters/api.yaml",
         listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
     )
 
@@ -77,6 +83,7 @@ tasks {
         dependsOn(generateKtorCode)
         dependsOn(generateKtorAuthCode)
         dependsOn(generateKtorInstantDateTimeCode)
+        dependsOn(generateKtorQueryParametersCode)
     }
 
 
--- a/end2end-tests/ktor/src/test/kotlin/com/cjbooms/fabrikt/servers/ktor/KtorInstantDateTimeTest.kt
+++ b/end2end-tests/ktor/src/test/kotlin/com/cjbooms/fabrikt/servers/ktor/KtorInstantDateTimeTest.kt
@@ -47,11 +47,11 @@ class KtorInstantDateTimeTest {
                 instantDateTimeRoutes(InstantDateTimeControllerImpl(listCapturingSlot, param2CapturingSlot))
             }
 
-            val response = client.get("/instant-date-time?explode_list_query_param=2025-02-16T10:52:46Z&explode_list_query_param=2025-02-16T11:52:46Z")
+            val response = client.get("/instant-date-time?explode_list_query_param=2025-02-16T10:52:46Z&explode_list_query_param=2025-02-16T11:52:46Z&query_param2=2025-02-16T10:52:46Z")
 
             assertEquals(HttpStatusCode.OK, response.status)
             assertEquals(listOf(Instant.parse("2025-02-16T10:52:46Z"), Instant.parse("2025-02-16T11:52:46Z")), listCapturingSlot.captured)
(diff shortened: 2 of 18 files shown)


ulrikandersen (commented):
First of all; thanks for the contribution!

I am curios to know what's preventing you from also handling required parameters? 
Handling non-required only feels a bit indeterministic from a user's perspective. I think a user would expect the same conversion to happen regardless of required/optional.

nsprod:
> First of all; thanks for the contribution!
> 
> I am curios to know what's preventing you from also handling required parameters? Handling non-required only feels a bit indeterministic from a user's perspective. I think a user would expect the same conversion to happen regardless of required/optional.

Agreed, it is not optimal. I focused on optional parameters mainly because this is the use case I currently have at work and it would unblock my team. But I'm going to add support for required parameters 👍

ulrikandersen:
> Agreed, it is not optimal. I focused on optional parameters mainly because this is the use case I currently have at work and it would unblock my team. And I'm new to Kotlin / Ktor so I wanted to get a first feedback on the changes I have made. But I'm going to add support for required parameters 👍

The changes you made so far are good. That's also why I am encouraging you to handle the required ones too 🙂

For tests it is great to reuse the `instantDateTime` example, and in addition I think it would be useful with an example that demonstrates the different cases of (query) parameters that you are adding code for (string with format, required/optional). This example can later be extended to cover more cases. Makes sense?

For handling required fields perhaps you could add a `getTypedOrFail` that we can use instead of `io.ktor.server.util.getOrFail`.

end2end-tests/ktor/build.gradle.kts
```diff
@@ -62,10 +66,24 @@ tasks {
...
+    val generateKtorInstantDateTimeCode = createCodeGenerationTask(
+        "generateKtorInstantDateTimeCode",
+        "src/test/resources/examples/instantDateTime/api.yaml",
+        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
+    )
+
+    val generateKtorQueryParametersCode = createCodeGenerationTask(
+        "generateKtorQueryParametersCode",
+        "src/test/resources/examples/queryParameters/api.yaml",
+        listOf("--type-overrides", "DATETIME_AS_INSTANT", "--serialization-library", "KOTLINX_SERIALIZATION")
```
ulrikandersen, line 78: The type override has not effect when Kotlinx Serialization is used and can be removed: It is always `kotlinx.datetime.Instant` ([source](https://github.com/cjbooms/fabrikt/blob/master/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinTypeInfo.kt#L92)).

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
```diff
@@ -494,3 +545,29 @@ private data class IncomingParametersByType(
 )
 
 private fun TypeName.isUnit(): Boolean = this == Unit::class.asTypeName()
+
+private fun RequestParameter.requiresKtorDataConversionPlugin(): Boolean {
```
ulrikandersen, line 549: I'm afraid the decision on when to use `DataConversionPlugin` is a be a bit more tricky than this. At first glance `DefaultConversionService` [handles only primitive types](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/common/src/io/ktor/util/converters/ConversionService.kt#L48), but the service hands off the decision further to platform specific implementations like [this one for JVM](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt#L11) where more types are handled by default (enums, BigDecimal, UUID, ...).

Perhaps we need to explicitly whitelist types that we wish to handle with `DataConversionPlugin` instead?

nsprod: Nice catch! 

Actually it might be even easier. I think we can forget about the `DefaultConversionService` and always call `call.application.conversionService`. As it seems the `DataConversion` class first tries to use a custom converter if any support a given type and if it doesn't it fallbacks on using `DefaultConversionService`. ([source](https://github.com/ktorio/ktor/blob/main/ktor-utils/common/src/io/ktor/util/converters/DataConversion.kt#L24))

```kotlin
override fun fromValues(values: List<String>, type: TypeInfo): Any? {
        if (values.isEmpty()) {
            return null
        }
        val converter = converters[type.type] ?: DefaultConversionService
        return converter.fromValues(values, type)
    }
```

So I can remove `requiresKtorDataConversionPlugin` altogether and always use `call.application.conversionService`.
If you don't install the `DataConversion`plugin, Ktor `DataConversion` class will always fallback on using the `DefaultConversionService`.
If you install the `DataConversion`plugin, Ktor `DataConversion` class will either use your custom converter if it matches the type being converted or fallback on using the `DefaultConversionService` if the type doesn't match.

Does this sound correct to you?

ulrikandersen: Unfortunately `call.application.conversionService` seems to be an extension function defined in the DataConversion plugin ([source](https://github.com/ktorio/ktor/blob/03b356b3389d2823eaa82d439b308e8fbc967955/ktor-server/ktor-server-plugins/ktor-server-data-conversion/common/src/io/ktor/server/plugins/dataconversion/DataConversion.kt#L37)) meaning that the generated code will not compile unless that dependency has been added to the project.

After having though more about this, I am learning towards your current implementation and relying on the `DataConversion` plugin for anything not primitive. 

The [JVM specific](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt#L11) DefaultConversionService has support for a few additional types (`BigDecimal`, `BigInteger` and `UUID`), but the user can really easily provide conversion for those using the plugin. This approach also solves an issue I have observed with enums: the default conversion service [compares on the enum name](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt#L18) where we actually want it to compare on the the enum value. Example from your spec: `?order=ASC` would work with the default converter, but `?order=asc` (which is the casing used in the spec) would not.

src/test/resources/examples/queryParameters/api.yaml
```diff
@@ -0,0 +1,42 @@
...
+      parameters:
+        - $ref: "#/components/parameters/PrimitiveTypeQueryParam"
+        - $ref: "#/components/parameters/EnumQueryParam"
+      responses:
+        200:
+          description: "successful operation"
+          content:
+            application/vnd.custom.media+json:
+              schema:
+                $ref: "#/components/schemas/QueryResult"
```
ulrikandersen, line 18: Super minor thing, but avoids the `TypedApplicationCall<Any>` in the generated code.
```suggestion
                $ref: "#/components/schemas/QueryParamsResult"
```

ulrikandersen (commented):
You were quicker than me 🙈 See my comment above.

src/main/kotlin/com/cjbooms/fabrikt/generators/controller/KtorControllerInterfaceGenerator.kt
nsprod: > Unfortunately call.application.conversionService seems to be an extension function defined in the DataConversion plugin ([source](https://github.com/ktorio/ktor/blob/03b356b3389d2823eaa82d439b308e8fbc967955/ktor-server/ktor-server-plugins/ktor-server-data-conversion/common/src/io/ktor/server/plugins/dataconversion/DataConversion.kt#L37)) meaning that the generated code will not compile unless that dependency has been added to the project.

Right, this is exactly why I introduced `requiresKtorDataConversionPlugin`. I was tired yesterday haha.

> After having though more about this, I am learning towards your current implementation and relying on the DataConversion plugin for anything not primitive.

The [JVM specific](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt#L11) DefaultConversionService has support for a few additional types (BigDecimal, BigInteger and UUID), but the user can really easily provide conversion for those using the plugin. This approach also solves an issue I have observed with enums: the default conversion service [compares on the enum name](https://github.com/ktorio/ktor/blob/e63cf88c29d60f76495ee1aa26c73a33fe9b85af/ktor-utils/jvm/src/io/ktor/util/converters/ConversionServiceJvm.kt#L18) where we actually want it to compare on the the enum value. Example from your spec: ?order=ASC would work with the default converter, but ?order=asc (which is the casing used in the spec) would not.

Agreed, given the kind of types you usually deal with in query parameters, this should be pretty straightforward to provide a custom implementation if necessary.

I'm going to roll back my last commit.

PS: One other solution would be to rely on reflection to test if the `DataConversion` dependency is part of the project. I'm not a seasoned JVM developer, but I guess we don't want to go this way.

src/test/resources/examples/queryParameters/api.yaml
nsprod: Nice catch 👍

end2end-tests/ktor/src/test/kotlin/com/cjbooms/fabrikt/servers/ktor/KtorInstantDateTimeTest.kt
```diff
@@ -0,0 +1,115 @@
...
+        testApplication {
+            configure(installInstantConverter = true)
+
+            routing {
+                instantDateTimeRoutes(InstantDateTimeControllerImpl(listCapturingSlot, param2CapturingSlot))
+            }
+
+            val response = client.get("/instant-date-time?query_param2=20-02-16T10:52:46Z")
+
+            assertEquals(HttpStatusCode.BadRequest, response.status)
```
ulrikandersen, line 72: Would we good to assert on the response message as well so we know that it is the correct reason for a 400. 

You can use the StatusPages plugin:

```kotlin
install(StatusPages) {
    exception<BadRequestException> { call, cause ->
        call.respond(HttpStatusCode.BadRequest, cause.message ?: "Bad Request")
    }
}
```

ulrikandersen (approved):
Nice work 💪 

I think we should extend this to apply to path parameters too, but we can do that in a separate future PR.

end2end-tests/ktor/src/test/kotlin/com/cjbooms/fabrikt/servers/ktor/KtorInstantDateTimeTest.kt
ulrikandersen: Fixed in https://github.com/cjbooms/fabrikt/pull/376.

nsprod: Thanks for the fix!

PR #378, opened 2025-02-22
Fix inheritance of array properties

mkostrzeba:
Properties of the array type that contain non-inlined objects/enums are not inherited correctly in the current implementation - generic parameter is prefixed with the name of the enclosing class, e.g. instead of `List&lt;Pet&gt;` we have `List&lt;EnclosingTypePet&gt;`


cjbooms:
Nice work. Thanks for the fix.

cjbooms:
Thanks for the contribution. I have released this now.

Issue #379, opened 2025-02-25
Old version of guava is causing conflicts


ethanmdavidson:
While trying to use the gradle plugin, I came across an issue caused by fabrikt having a transitive dependency on an old version of guava. See https://github.com/acanda/fabrikt-gradle-plugin/issues/96

fabrikt has a transitive dependency on guava 18:

```
> ./gradlew dependencies --configuration=runtimeClasspath

<snip>

+--- com.reprezen.kaizen:openapi-parser:4.0.4
|    +--- com.reprezen.jsonoverlay:jsonoverlay:4.0.4
|    |    \--- org.eclipse.xtend:org.eclipse.xtend.lib:2.11.0
|    |         +--- org.eclipse.xtext:org.eclipse.xtext.xbase.lib:2.11.0
|    |         |    \--- com.google.guava:guava:[14.0,19) -> 18.0
```

One way I've solved this before is by [relocating the package](https://gradleup.com/shadow/configuration/relocation/) so that both versions can coexist in the classpath. This is probably the most straightforward solution.

Another option is to upgrade `org.eclipse.xtend.lib` to [the latest version](https://mvnrepository.com/artifact/org.eclipse.xtend/org.eclipse.xtend.lib/2.37.0), which depends on guava 33. This option doesn't seem so good because the reprezen dependencies seem to be unmaintained, and would probably need to be forked. There is [a kotlin fork of the reprezen stuff](https://github.com/wakaztahir/K-OpenApi-Parser) which could be tried too. This is probably a better long-term plan, but not something I can commit to at this point.

I'll start putting together a PR for the package relocation, but wanted to open this issue for discussion in case you want to take a different route.

cjbooms:
Causing conflicts where? Fabrikt should not be on any runtime class paths.

cjbooms:
No problem accepting PRs, was just curious. I'm following this on my phone so maybe missed some context. Will review PR later. Thanks for the contribution

PR #380, opened 2025-02-25
relocate old guava libs to prevent conflicts

ethanmdavidson:
Fixes https://github.com/cjbooms/fabrikt/issues/379

Confirmed by building fabrikt and plugin locally, and using the plugin in a build that used to trigger the conflict.


cjbooms:
Thanks for the contribution. I have released this now

PR #383, opened 2025-03-08
add springboot's http interface generator

mirror-kt:
SpringBoot's HTTP Interface is a declarative http client like Retrofit or OpenFeign.
https://docs.spring.io/spring-framework/reference/integration/rest-clients.html#rest-http-interface

Since SpringBoot has now removed OpenFeign from the official project and seems to be focusing on this HTTP Interface, we thought it reasonable to add support for it and implemented it.

This client supports ResponseEntity and suspend methods, so we move the relevant code to ClientGeneratorUtils and so on.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
@@ -125,7 +125,7 @@ object ClientGeneratorUtils {
     }
 
     /**
-     * Adds suspend as modified to the func spec so that i can be used with CoroutineFeign
+     * Add suspend as modified to method definitions on supported clients, ex. CoroutineFeign, Spring HTTP Interface.
      */
     fun FunSpec.Builder.addSuspendModifier(options: Set<ClientCodeGenOptionType>): FunSpec.Builder {
         if (options.contains(ClientCodeGenOptionType.SUSPEND_MODIFIER)) {
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
@@ -1,5 +1,6 @@
 package com.cjbooms.fabrikt.generators.client
 
+import com.cjbooms.fabrikt.cli.ClientCodeGenOptionType
 import com.cjbooms.fabrikt.configurations.Packages
 import com.cjbooms.fabrikt.generators.GeneratorUtils
 import com.cjbooms.fabrikt.generators.GeneratorUtils.getPrimaryContentMediaType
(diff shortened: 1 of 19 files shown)


src/main/kotlin/com/cjbooms/fabrikt/generators/client/SpringHttpInterfaceGenerator.kt
```diff
@@ -0,0 +1,273 @@
...
+            }
+            return resolvedResource
+        }
+    }
+
+    override fun generateLibrary(options: Set<ClientCodeGenOptionType>): Collection<GeneratedFile> = setOf(
+        generatedAnnotationFile()
+    )
+
+    private fun generatedAnnotationFile(): SimpleFile {
```
mirror-kt, line 260: The `@GeneratedClient` annotation is not required for the feature itself, but SpringBoot often has use cases where classes are searched for using annotations, and this is what it is for

src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
```diff
@@ -100,8 +111,33 @@ object ClientGeneratorUtils {
...
+                }
+            }
             builder.build()
         }
         return this.addParameters(specs)
     }
+
+    /**
+     * Adds suspend as modified to the func spec so that i can be used with CoroutineFeign
+     */
```
cjbooms, line 126: is this kdoc still valid? Is ti only for CoroutineFeign still. Remove doc entirely if not

src/main/kotlin/com/cjbooms/fabrikt/generators/client/SpringHttpInterfaceGenerator.kt
cjbooms: Where is this annotation defined? I cannot see any class imports

cjbooms:
Thanks for the contribution, happy to accept this. Few minor questions

src/main/kotlin/com/cjbooms/fabrikt/generators/client/SpringHttpInterfaceGenerator.kt
mirror-kt: It is generated in the same location as the client's interface.
Since they are in the same package, there should be no need for an import statement.

src/main/kotlin/com/cjbooms/fabrikt/generators/client/ClientGeneratorUtils.kt
mirror-kt: fixed e105d48

src/main/kotlin/com/cjbooms/fabrikt/generators/client/SpringHttpInterfaceGenerator.kt
cjbooms: I'm not so sure about this, it feels wrong to be generating a random annotation just in case it is useful. Could developers instead define their own annotation and add it to the implementation classes?

mirror-kt: For example, like creating an option like --http-client-annotations <annotation FQCN> and annotating the generated clients?

 It seems useful not only for this client type, so I'll do it in another PR and revert it in this one.

mirror-kt: Opened #386 and revert `@GeneratedClient` in this PR

PR #385, opened 2025-03-10
returns Unit when any successful response not exists

mirror-kt:
In the original implementation, if only 204 No Content and error system responses were defined in the response, the client return value was the model of the error with the response body.

This was considered undesirable behavior, so it has been modified to return Unit if none of the response bodies should be returned in a 2xx.


cjbooms:
👍

cjbooms:
LGTM - thanks for the addtion

Issue #387, opened 2025-04-04
uniqueItems Support - generate Set instead of List for unique items


AlexanderKomi:
Hi, 

i am using fabrikt to generate an array of values, with a unique constraint.

The OpenApi Spec 3.0 includes the property "uniqueItems" to specify if an array items should be unique.

Currently fabrikt generates a Kotlin.List for the array.
It would be helpful to generate a Kotlin.Set instead, if the "uniqueItems" property is set to "true".

---

Example from the Spec:
```yaml
type: array

items:

  type: integer

uniqueItems: true
```
[1, 2, 3] – valid
[1, 1, 3] – not valid
[ ] – valid

AlexanderKomi:
Currently I am not sure on how to do it.
However I would greatly appreciate the effort to implement this feature.

AlexanderKomi:
Fixed the generation for Sets.
Please review the changes.

PR #389, opened 2025-04-04
Unique items support

AlexanderKomi:
an attempt to to fix issue: https://github.com/cjbooms/fabrikt/issues/387


src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
```diff
@@ -68,6 +68,7 @@ class ModelGeneratorTest {
         "byteArrayStream",
         "untypedObject",
         "primitiveTypes",
+        "uniqueItems"
```
cjbooms, line 71: Rather than add a completely new test case and spec, just add an array with this property set to true on an existing test spec.

cjbooms:
Nice work, great little addition

src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
AlexanderKomi: I moved the test case to the array generation test cases.
Also I added a test case to check if an property with the "uniqueItems" flag does generate a Set.
Unfortunately this does not work and I have no idea why.

AlexanderKomi:
Unfortunately my changes do not have the intended effect. 
Currently it does not work and I have no idea why.
The tests fail at this point and I will leave it as it is.
Someone should feel free to change whatever is necessary to make it work.

AlexanderKomi:
I rebased on the master branch and fixed the implementation to generate sets.
@cjbooms Please take a look at the implementation and suggest changes, if any are required.
Not sure if the README.md should include this features aswell.

cjbooms:
Great, I should be able to get a release out this evening for you. nice work

Issue #390, opened 2025-04-18
openapi parser crashes when used with anthropic sdk openapi file


christophsturm:
anthropic does not publish an official openapi file but in their test suite they use an openapi file for mocking:
https://github.com/anthropics/anthropic-sdk-java/blob/main/.stats.yml#L2
and if i want to generate classes from this file i get a null pointer exception in the openapi parser:

```
java -jar build/libs/fabrikt-21.2.0-3-gf763a82.jar --api-file ../../ai/ailib/anthropic-api.yml --base-package blah 2>&1 | pbcopy
Apr. 18, 2025 5:41:39 PM com.cjbooms.fabrikt.cli.CodeGen generate
INFORMATION: Generating code and dumping to ./
Exception in thread "main" java.lang.IllegalArgumentException: The Kaizen openapi-parser library threw a NPE exception when parsing this API. This is commonly due to an external schema reference that is unresolvable, possibly due to a lack of internet connection
	at com.cjbooms.fabrikt.util.YamlUtils.parseOpenApi(YamlUtils.kt:62)
	at com.cjbooms.fabrikt.model.SourceApi.<init>(SourceApi.kt:35)
	at com.cjbooms.fabrikt.model.SourceApi$Companion.create(SourceApi.kt:31)
	at com.cjbooms.fabrikt.cli.CodeGen.generate(CodeGen.kt:85)
	at com.cjbooms.fabrikt.cli.CodeGen.main(CodeGen.kt:22)
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.CharSequence.length()" because "this.text" is null
	at java.base/java.util.regex.Matcher.getTextLength(Matcher.java:1769)
	at java.base/java.util.regex.Matcher.reset(Matcher.java:415)
	at java.base/java.util.regex.Matcher.<init>(Matcher.java:252)
	at java.base/java.util.regex.Pattern.matcher(Pattern.java:1144)
	at com.reprezen.kaizen.oasparser.val.ValidatorBase.checkPattern(ValidatorBase.java:99)
	at com.reprezen.kaizen.oasparser.val.ValidatorBase.validateStringField(ValidatorBase.java:89)
	at com.reprezen.kaizen.oasparser.val.ValidatorBase.validateStringField(ValidatorBase.java:79)
	at com.reprezen.kaizen.oasparser.val3.SchemaValidator.runObjectValidations(SchemaValidator.java:79)
[...]

PR #391, opened 2025-04-19
Fix NPE on empty type.

cjbooms:
The spec file provided in issue #390 has the following blank type that causes the kaizen parser to crash, so I have added a little function to fix this:
```
    Model:
      type:
      title: Model
```

Also improving the handling of 3.1 nullable definitions, by expanding support to include anyOf definitions. This was a further stumbling block with the API provided in #390


christophsturm:
i tried this pr with my model file and it works now!

PR #393, opened 2025-04-21
Feature/add annotations

cjbooms:
add option to specify additional annotations to controllers/clients/models.
- `--http-controller-class-additional-annotations`
- `--http-controller-method-additional-annotations`
- `--http-client-class-additional-annotations`
- `--http-client-method-additional-annotations`
- `--http-model-annotations`
- `x-additional-annotations`

This was created as a separate PR in response to review comments in #383 

`x-additional-annotations` is useful to annotate with `@Cacheable` when it is a cacheable API call, for example.


cjbooms:
@ulrikandersen / @averabaq any thoughts on this addition? I'm not sure I see the value, but also don't want to gate keep this feature if folks would find it beneficial.

mirror-kt:
I would like to have the ability to add arbitrary annotations, but this PR implementation only allows annotations with no arguments.
This is a major limitation, and would ideally take the form of an extension or plugin.

mirror-kt:
> would it not be possible to put the annotations on the implementing classes instead of the generated interfaces?

In the context of the Spring HTTP Interface, this is not feasible because the implementation class is dynamically generated by reflection.
The official documentation only explains how to annotate the interface

cjbooms:
> I would like to have the ability to add arbitrary annotations, but this PR implementation only allows annotations with no arguments.
> This is a major limitation, and would ideally take the form of an extension or plugin.

Thank you for the contribution @mirror-kt but the maintainers think this would be best left out of fabrikt. Users can annotate implementation classes directly, post generation. And as you say above, this feature would be a slippery slope towards accepting more complicated named attributes in the future. Best to instead keep the implementation tight to absolutely necessary annotations only. 

Again, thanks for the contribution. Hopefully the last feature you added is working well for you

Issue #392, opened 2025-04-21
error message when used with openai sdk openapi file


christophsturm:
i am playing arount with building an llm client lib with fabrikt so i tried running fabrikt on the openai openapi file and i get an error message.

here is the file:
https://github.com/openai/openai-openapi/blob/master/openapi.yaml

```
% java -jar build/libs/fabrikt-21.2.0-4-g10b53c7.jar --api-file ../../ai/ailib/apis/openai-api.yml --base-package generated.openai

Exception in thread "main" java.lang.IllegalStateException: Unknown OAS type: string and format: null and specialization: UNKNOWN_ADDITIONAL_PROPERTIES for key: additionalPropertiesValue
        at com.cjbooms.fabrikt.model.OasType$Companion.toOasType(OasType.kt:76)
        at com.cjbooms.fabrikt.model.KotlinTypeInfo$Companion.from(KotlinTypeInfo.kt:83)
        at com.cjbooms.fabrikt.model.KotlinTypeInfo$Companion.from(KotlinTypeInfo.kt:132)
        at com.cjbooms.fabrikt.model.KotlinTypeInfo$Companion.from$default(KotlinTypeInfo.kt:75)
        at com.cjbooms.fabrikt.model.SchemaInfo.<init>(SourceApi.kt:14)
        at com.cjbooms.fabrikt.model.SourceApi.<init>(SourceApi.kt:45)
        at com.cjbooms.fabrikt.model.SourceApi$Companion.create(SourceApi.kt:31)
        at com.cjbooms.fabrikt.cli.CodeGen.generate(CodeGen.kt:85)
        at com.cjbooms.fabrikt.cli.CodeGen.main(CodeGen.kt:22)
```
but the interesting thing is, that file does not even contain the string "additionalPropertiesValue"

that is the branch with the anthropic fixes.
in the master branch it looks like this:
```
Exception in thread "main" java.lang.IllegalStateException: Unknown OAS type: string and format: null and specialization: UNKNOWN_ADDITIONAL_PROPERTIES
```

christophsturm:
The openai file does now no longer create an error message, but the generated code does not compile.
i use these args:
```
                args = listOf(
                    "--output-directory", generationDir,
                    "--base-package", "model.$packageName",
                    "--api-file", apiFile,
                    "--targets", "http_models",
                    "--validation-library", "no_validation",
                )
```
and one of the files that is generated is this:
```
package model.openai.models

[...]

PR #403, opened 2025-04-24
Fix inline handling of aggregations under arrays

cjbooms:
Fix the issue where aggregations inlined under arrays were referencing the incorrectly generated class name.

In the end this was a very small fix. When within an anyOf or allOf, we need to check that the enclosing schema we are about to chose is not itself an inlined definition, and if it is we revert to the outer schema.
from this:
```
enclosingSchema =  this
```
to
```
enclosingSchema = if (this.isInlinedObjectDefinition()) enclosingSchema else this
```

Closes https://github.com/cjbooms/fabrikt/issues/392


src/test/resources/examples/inlinedAggregatedObjects/api.yaml
```diff
@@ -8,31 +8,60 @@ paths: { }
...
+            - $ref: "#/components/schemas/SimpleObjOne"
+        aggregationOfMany:
           allOf:
-            - $ref: "#/components/schemas/Person"
-            - $ref: "#/components/schemas/Company"
-
-    Person:
+            - $ref: "#/components/schemas/SimpleObjOne"
+            - $ref: "#/components/schemas/SimpleObjTwo"
+        arrayWithAllOfAggregationOfMany:
```
cjbooms, line 21: This reproduced the issue described here, where the generated classes were referencing non-existent types
https://github.com/cjbooms/fabrikt/issues/392#issuecomment-2821255513

PR #420, opened 2025-06-16
Update version of fabrikt-gradle-plugin in README.md to 1.16.1

acanda:
I updated the gradle plugin to Fabrikt 22.5.0.


DreamEmulator:
#420! Perfect moment to discover this library..
Thanks for spamming Reddit with Fabrikt.
Was stunned how sloppy the Kotlin OAS client was compared to Apple/Swift.

This makes good on Apple tho, and then some.

Keep up the great work! 



And sorry for hijacking the comments.

PR #421, opened 2025-06-18
Rename 'client' variable so it doesn't clash with other params

davwil00:
If using a query param/header called `client` the code fails to compile due to a clash of names/types. Renaming the `client` to `okHttpClient` should mean there's a much smaller risk of clashes.


cjbooms:
Thanks for the contribution. This is released now

PR #442, opened 2025-09-03
Support JSON encoded header objects

davwil00:
Currently when headers with complex types are used, they're generated with an `Any` type which then doesn't compile when the client tries to set them (it expects a string).

For headers that have a content type of `application/json` and define the object type in the schema, it looks like they could relatively easily be exposed with the correct type and converted to a string with the existing object mapper.

Here's an attempt to do that along with a new test.


cjbooms:
Thanks @davwil00. I'm wondering though is the bug here that it should be String rather than Any, instead of trying to do a complex type?

davwil00:
#449 fixes the immediate issue, thanks!

PR #444, opened 2025-09-07
fix: default value for floating numbers

### LINKED ISSUE #443: Floating numbers with a default value does not compile
Here is an example of schema  + generated code that does not work:
```yaml
FloatWithDefaultValue:
  type: object
  properties:
    float_with_default:
      type: number
      format: float
      default: 0
    double_with_default:
      type: number
      format: double
      default: 0
```

This will generate the following Kotlin code:
```kotlin
public data class FloatWithDefaultValue(
  @param:JsonProperty("float_with_default")
  @get:JsonProperty("float_with_default")
  @get:NotNull
  public val floatWithDefault: Float = 0,
  @param:JsonProperty("double_with_default")
  @get:JsonProperty("double_with_default")
  @get:NotNull
  public val doubleWithDefault: Double = 0,
)
```

It does not compile because `0` is considered as an `Int`.
```
> Task :end2end-tests:models-jackson:compileKotlin FAILED
e: file:///Users/nathanfallet/git/opensource/fabrikt/end2end-tests/models-jackson/build/generated/src/main/kotlin/com/example/models/FloatWithDefaultValue.kt:12:40 Initializer type mismatch: expected 'kotlin.Float', actual 'kotlin.Int'.
e: file:///Users/nathanfallet/git/opensource/fabrikt/end2end-tests/models-jackson/build/generated/src/main/kotlin/com/example/models/FloatWithDefaultValue.kt:16:42 Initializer type mismatch: expected 'kotlin.Double', actual 'kotlin.Int'.
```

So we should write it as `0.0` instead. This applies to any default value that is written as an integer (could be `1`, `2`, ... but not `1.2` for example)

I'm making a PR to fix this right now (just opening the issue so the problem is fully explained and tracked)

nathanfallet:
Fixes #443 

I first added the schema in `api.yaml` to see the issue (the one mentioned in #443), and then I added the suffix part to the default value to fix it.

### DIFF
--- a/end2end-tests/models-jackson/openapi/api.yaml
+++ b/end2end-tests/models-jackson/openapi/api.yaml
@@ -431,3 +431,23 @@ components:
         properties:
           bar_prop:
             type: string
+
+    FloatWithDefaultValue:
+      type: object
+      properties:
+        float_with_default_integer:
+          type: number
+          format: float
+          default: 0
+        double_with_default_integer:
+          type: number
+          format: double
+          default: 0
+        float_with_default_float:
+          type: number
+          format: float
+          default: 0.1
+        double_with_default_float:
+          type: number
+          format: double
+          default: 0.1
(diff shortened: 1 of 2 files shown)


ulrikandersen:
I was expecting to see added/changed test models?

cjbooms:
> > I'm just expecting the generated model to compile. You can try the same schema with the previous version and see the project does not build.
> 
> Ah yes I see its part of the end2end tests 👍
> 
> Can you add it to the regular unit tests too? It is nice to to be able to see the expected output directly for a given OpenAPI.

Agreed. I just added one direct and merged this.

Thanks for the contribution @nathanfallet  💪

PR #451, opened 2025-09-17
Improve test setup.

cjbooms:
Ensure that all files in expected directory are generated. Previously only validated that the ones generated were found in expected directory, so missing files were not alarmed on.

This code change detected files that should not have been present in test folders and this PR removes them.


src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
```diff
@@ -121,14 +122,18 @@ class ModelGeneratorTest {
             ?.let(::readFolder)
             ?: emptyMap()
         tempFolderContents.forEach {
-            if (expectedModels.containsKey(it.key)) {
-                assertThat((it.value))
-                    .describedAs("expected model '${it.key}' does not match the given value.")
-                    .isEqualTo(expectedModels[it.key])
+            if (expectedModels.contains(it.key)) {
+                assertThatGenerated(it.value)
```
cjbooms, line 126: this will allow the auto-correction to be turned on if there are mass file changes required in future.

PR #456, opened 2025-10-03
Add `@Contextual` annotations to array elements for KOTLINX_SERIALIZATION

ulrikandersen:
Types like UUID, URI, BigDecimal, ByteArray, Instant, and LocalDateTime require `@Contextual` annotation with kotlinx.serialization. Previously, these annotations were only added to singular properties, not array elements, causing serialization issues at runtime.

Now generates: `List<@Contextual UUID>` instead of: `List<UUID>` 👌


cjbooms (approved):
LGTM. You were on fire on Friday!

PR #469, opened 2025-10-30
Add support to okhttp client code generation for getting raw byte content

averabaq:
Add support to okhttp client code generation for getting raw byte contntent on `application/octet-stream` response types.

This previously generated code for deserializing non-json byte array content:
```
    @Throws(ApiException::class)
    public fun postBinaryData(
        applicationOctetStream: ByteArray,
        additionalHeaders: Map<String, String> = emptyMap(),
        additionalQueryParameters: Map<String, String> = emptyMap(),
    ): ApiResponse<ByteArray> { 
       ... 
      return request.execute(okHttpClient, objectMapper, jacksonTypeRef())
}
```

Now we allow to deserialize raw bytes in-memory:

```
    @Throws(ApiException::class)
    public fun postBinaryData(
        applicationOctetStream: ByteArray,
        additionalHeaders: Map<String, String> = emptyMap(),
        additionalQueryParameters: Map<String, String> = emptyMap(),
    ): ApiResponse<ByteArray> { 
       ... 
      return request.execute(okHttpClient)
}
```

### DIFF
--- a/src/main/resources/templates/client-code/http-util.kt.hbs
+++ b/src/main/resources/templates/client-code/http-util.kt.hbs
@@ -72,21 +72,6 @@ fun ResponseBody.deserialize(): ByteArray? = this.byteStream().readAllBytes()
 
 fun String?.isNotBlankOrNull() = if (this.isNullOrBlank()) null else this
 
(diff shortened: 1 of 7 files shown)


averabaq:
> Welcome back Alejandro!

I've never left 🙂

cjbooms (approved):

PR #476, opened 2025-11-09
Make polymorphic allOf detection more targeted.

### LINKED ISSUE #475: oneOf generation inside of arrays still broken
I had to downgrade my version of Fabrikt to 21.2.0 to fix this.  That is the last version that will successfully generate my schema.  I intend to investigate this later unless someone else knows the exact nature of the issue at present. This PR was originally intended to fix this problem, but it seems to still have edge cases that do not generate correctly: https://github.com/fabrikt-io/fabrikt/pull/401

I think this may have something to do with my example's `AModel` and `BModel` being used as oneOf fields with the `CModel` being a required subfield of each.  The generated class that fails comes out being named: `public val a: List<ExampleResponseCModel>`, which smashes together the name of the top-level response class and the `CModel` name, which isn't actually a class that is generated.  The proper generation in version 21.2.0 just uses `CModel` as the output value, which seems to be the correct thing to do.  With sealed interfaces, the expected `ExampleResponseCModel` sealed interface _should_ be being generated, but it isn't even when providing that model generation option.  However, even with the sealed interface generation option disabled, it still tries to point to the nonexistent sealed interface name.

Here are examples of my schema (redacted) that describe the issue:

Source model with bad generation:
```json
      "ExampleResponse": {
        "required": [
          "a",
          "b",
          "timestamp"
        ],
        "type": "object",
        "properties": {
          "timestamp": {
            "type": "string",
            "format": "date-time"
          },
          "a": {
            "title": "<redacted>",
            "type": "array",
            "items": {
              "oneOf": [
                {
                  "$ref": "#/components/schemas/AModel"
                },
                {
                  "$ref": "#/components/schemas/BModel"
                }
              ]
            }
          },
          "b": {
            "title": "<redacted>",
            "type": "array",
            "items": {
              "oneOf": [
                {
                  "$ref": "#/components/schemas/AModel"
                },
                {
                  "$ref": "#/components/schemas/BModel"
                }
              ]
            }
          }
        }
      },
```

AModel example:
```json
      "AModel": {
        "required": [
          "a",
          "b",
          "c"
        ],
        "type": "object",
        "allOf": [
          {
            "$ref": "#/components/schemas/CModel"
          },
          {
            "type": "object",
            "properties": {
              "b": {
                "title": "<redacted>",
                "type": "number"
              },
              "c": {
                "title": "<redacted>",
                "type": "number"
              }
            }
          }
        ]
      },
```

BModel example:
```json
      "BModel": {
        "required": [
          "a",
          "b"
        ],
        "type": "object",
        "allOf": [
          {
            "$ref": "#/components/schemas/CModel"
          },
          {
            "type": "object",
            "properties": {
              "b": {
                "title": "<redacted>",
                "type": "number"
              }
            }
          }
        ]
      },
```

CModel example:
```json
      "CModel": {
        "title": "<redacted>",
        "required": [
          "a",
          "b"
        ],
        "type": "object",
        "properties": {
          "b": {
            "type": "number"
          },
          "a": {
            "title": "<redacted>",
            "type": "number"
          }
        }
      },
```

cjbooms:
We cannot just assume that if the first schema in a oneOf contains an allOf, that all the oneOf schemas will also contain that same allOf schema. This logic is a hangover from the very first attempts by fabrikt to handle polymorphism, long before the recommendation to use discriminated oneOf.

We're still not 100% bullet-proof with this logic, but this improves it by ensuring its firing is much more targeted.

Closes #475

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
@@ -287,8 +287,13 @@ object KaizenParserExtensions {
     private fun List<Schema>?.hasAnyDefinedProperties(): Boolean =
         this?.any { it.properties?.isNotEmpty() == true } == true
 
-    fun Schema.isOneOfPolymorphicTypes() =
-        this.oneOfSchemas?.firstOrNull()?.allOfSchemas?.firstOrNull() != null
+    fun Schema.isOneOfPolymorphicTypes(): Boolean {
+        val maybeAllOfInFirstOneOf = this.oneOfSchemas?.firstOrNull()?.allOfSchemas?.firstOrNull()
+        return if (maybeAllOfInFirstOneOf != null) {
+            this.oneOfSchemas.all { it.allOfSchemas.contains(maybeAllOfInFirstOneOf) }
+        } else false
+    }
+
 
     fun Schema.isInlinedOneOfSuperInterface() = isOneOfSuperInterfaceOnly() && isInlinedPropertySchema()
 
--- a/src/test/resources/examples/inlinedAggregatedObjects/api.yaml
+++ b/src/test/resources/examples/inlinedAggregatedObjects/api.yaml
@@ -38,6 +38,28 @@ components:
                   id:
                     type: string
                     description: The identifier of the chat message.
+        arrayWithOneOf:
+          type: array
+          items:
+            oneOf:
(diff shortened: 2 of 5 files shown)


ulrikandersen (approved):
Looks good!

Certainly fixes the immediate issue where this is currently resulting in `public val arrayWithOneOf: List<ContainerBModel>? = null` where `ContainerBModel` is a non-existing model 😬

PR #481, opened 2025-11-25
Upgrade to gradle 9.0

y-marion:
This updates the project to gradle version 9.0.

Changed the coordinates of the shadow plugin, as maintenance was move to gradleup. Did not upgrade to version 9.x of the shadow plugin directly, as i had not yet the time to review changes needed (as there are a lot of deprecations / removals in the shadow plugin).

This made it necessary to include the junit-platform-launcher as `testRuntimeOnly`. This is needed after the deprecation of the inclusion in gradle 8.2 and subsequent removal with 9.0 (see also: https://github.com/gradle/gradle/issues/34512 ).

### DIFF
--- a/end2end-tests/ktor-client-kotlinx/build.gradle.kts
+++ b/end2end-tests/ktor-client-kotlinx/build.gradle.kts
@@ -38,7 +38,6 @@ dependencies {
     testImplementation("io.ktor:ktor-server-test-host:$ktorVersion")
     testImplementation("io.mockk:mockk:1.13.7")
 
-    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
     testRuntimeOnly("org.junit.platform:junit-platform-launcher")
     testImplementation("org.junit.jupiter:junit-jupiter-api:$junitVersion")
     testImplementation("org.junit.jupiter:junit-jupiter-engine:$junitVersion")


build.gradle.kts
```diff
@@ -62,6 +62,8 @@ dependencies {
 
     implementation("org.jetbrains.kotlinx:kotlinx-datetime:$kotlinxDateTimeVersion")
 
+    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
```
cjbooms:
```suggestion
```

end2end-tests/models-jackson/build.gradle.kts
```diff
@@ -30,6 +30,8 @@ dependencies {
     implementation("com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion")
     implementation("com.fasterxml.jackson.datatype:jackson-datatype-jsr310:$jacksonVersion")
 
+    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
```
cjbooms:
```suggestion
```

end2end-tests/okhttp/build.gradle.kts
```diff
@@ -32,6 +32,8 @@ dependencies {
     implementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion")
     implementation("com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion")
 
+    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
```
cjbooms:
```suggestion
```

end2end-tests/openfeign/build.gradle.kts
```diff
@@ -34,6 +34,8 @@ dependencies {
     implementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion")
     implementation("com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion")
 
+    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
```
cjbooms:
```suggestion
```

end2end-tests/spring-http-interface/build.gradle.kts
```diff
@@ -32,6 +32,8 @@ dependencies {
     implementation("com.fasterxml.jackson.core:jackson-core:$jacksonVersion")
     implementation("com.fasterxml.jackson.core:jackson-annotations:$jacksonVersion")
 
+    // Junit-Platform-Launcher is needed for gradle 9.x compatibility
```
cjbooms:
```suggestion
```

y-marion:
> LGTM. Probably no need for the comments on top of the new junit dep though

i added them for completeness, but i see you already removed them.
I also looked into upgrading jackson (because we are using cyclonedx internally, and this causes a problem when running fabrikt from a gradle plugin, as the cyclonedx plugin requires jackson >= 2.19). Do you see any issue with bumping the version there? - i would create a new MR for this anyway

cjbooms:
> > LGTM. Probably no need for the comments on top of the new junit dep though
> 
> i added them for completeness, but i see you already removed them. I also looked into upgrading jackson (because we are using cyclonedx internally, and this causes a problem when running fabrikt from a gradle plugin, as the cyclonedx plugin requires jackson >= 2.19). Do you see any issue with bumping the version there? - i would create a new MR for this anyway

No issue. Any lib version bumps you can do is appreciated

PR #484, opened 2025-12-03
Fix Docker build and publish with Gradle 9

cjbooms:
Adds test sources to docker file for docker build

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,6 +12,8 @@ COPY .git ./.git
 
 # Copy source code
 COPY src ./src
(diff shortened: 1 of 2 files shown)


ulrikandersen:
> I haven't really looked at how that part of the build works, I just assumed that would result in a bloated docker image

Not a problem. It will be excluded from the resulting image:

```
# Copy the built jar from builder stage
COPY --from=builder /build/build/libs/fabrikt-*.jar /app/fabrikt.jar
```

PR #501, opened 2026-01-16
Generation Bug With OneOf containing allOf

cjbooms:
The older allOf polymorphism in fabrikt was breaking generation for cases where a oneOf contained references to objects that all contained an allOf.

This change further tightens the spots where that old allOf polymorphism logic gets triggered


src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
```diff
@@ -287,13 +291,21 @@ object KaizenParserExtensions {
     private fun List<Schema>?.hasAnyDefinedProperties(): Boolean =
         this?.any { it.properties?.isNotEmpty() == true } == true
 
-    fun Schema.isOneOfPolymorphicTypes(): Boolean {
+    fun Schema.isOneOfWhereAllTypesInheritFromACommonAllOfSuperType(): Boolean {
```
cjbooms: this name is more descriptive of what is actually being checked

src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
```diff
@@ -239,7 +242,8 @@ object KaizenParserExtensions {
 
     fun Schema.safeName(): String =
         when {
-            isOneOfPolymorphicTypes() -> this.oneOfSchemas.first().allOfSchemas.first().safeName()
+            isOneOfWhereAllTypesInheritFromACommonAllOfSuperType() && !isSealedInterfacesForOneOfEnabled() ->
```
cjbooms: Makes this logic dumber. In most cases we want to just use the class name, the case where we want to be smart and try and detect a better class is very narrow.

src/test/kotlin/com/cjbooms/fabrikt/util/GeneratedCodeAsserter.kt
```diff
@@ -13,8 +14,20 @@ class GeneratedCodeAsserter(val generatedCode: String) {
     companion object {
         // Set this to true to overwrite the expected files with the generated code when they don't match
         const val SHOULD_OVERWRITE_EXAMPLES = false
+        const val SHOULD_SKIP_ERRORS = false
 
         fun assertThatGenerated(generatedCode: String): GeneratedCodeAsserter = GeneratedCodeAsserter(generatedCode)
+
+        fun failGenerated(generatedCode: String) = GeneratedCodeAsserter(generatedCode)
+
+        fun maybeGenerateMissingFile(resourcePath: String, generatedCode: String) {
```
cjbooms: Unrelated change, but it's very handy to be able to generate all the test classes when you are experimenting with a new API spec

PR #506, opened 2026-01-17
Fix bug with inlined allOf as Items schema in array

cjbooms:
Fixes issues where parameterised type of an array was wrong when the items schema is an inlined allOf definition. Such as in below example

```
    ArrayWithInlinedAllOf:
      type: object
      properties:
        items:
          type: array
          items:
            allOf:
              - $ref: '#/components/schemas/ItemId'
              - type: object
                properties:
                  status_code:
                    type: integer
    ItemId:
      type: object
      properties:
        id:
          type: string
```

Was generating this:
```
public data class ArrayWithInlinedAllOf(
  @param:JsonProperty("items")
  @get:JsonProperty("items")
  @get:Valid
  public val items: List<ArrayWithInlinedAllOfArrayWithInlinedAllOf>? = null,
)
```

Fixed to correctly generate this
```
public data class ArrayWithInlinedAllOf(
  @param:JsonProperty("items")
  @get:JsonProperty("items")
  @get:Valid
  public val items: List<ArrayWithInlinedAllOfItems>? = null,
)
```

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinTypeInfo.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinTypeInfo.kt
@@ -186,9 +186,12 @@ sealed class KotlinTypeInfo(val modelKClass: KClass<*>, val generatedModelClassN
             enclosingSchema: Schema?,
             oasKey: String
         ): KotlinTypeInfo {
-            return if (arraySchema.itemsSchema.isInlinedObjectDefinition() && !arraySchema.itemsSchema.isUnsupportedComplexInlinedDefinition())
-                Object(ModelNameRegistry.getOrRegister(arraySchema, enclosingSchema))
-            else from(arraySchema.itemsSchema, oasKey, enclosingSchema)
+            return when {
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/cjbooms/fabrikt/model/KotlinTypeInfo.kt
```diff
@@ -184,6 +181,19 @@ sealed class KotlinTypeInfo(val modelKClass: KClass<*>, val generatedModelClassN
...
 
+        private fun getParameterizedTypeForArray(
+            arraySchema: Schema,
+            enclosingSchema: Schema?,
+            oasKey: String
+        ): KotlinTypeInfo {
+            return when {
+                arraySchema.itemsSchema.isInlinedObjectDefinition() && !arraySchema.itemsSchema.isUnsupportedComplexInlinedDefinition() ->
+                    Object(ModelNameRegistry.getOrRegister(arraySchema, enclosingSchema))
+                arraySchema.itemsSchema.isNotDefined() -> AnyType
```
cjbooms: Previosuly we threw an exception for this. But I think AnyType is a much nicer experience

PR #516, opened 2026-01-29
Implement ApiConfiguration for Ktor client

vekunz:
I would like to suggest a few extensions to the Fabrikt project. This Pull Request adds an optional parameter `apiConfiguration: ApiConfiguration` to all generated client functions of the Ktor client. I oriented myself on the OpenAPI Generator Client for Swift when implementing.

ApiConfiguration has two features (but it could be extended in the future).

- It passes the server URL from the `openapi.yml` to the client functions.
- It allows adding custom headers to the HTTP call

**Example for server URL**

The reason for this is pretty obvious. When the openapi.yml already contains a basePath, I want to use it without copying it.

```yaml
openapi: 3.1.0

servers:
  - url: https://domain.com

paths:
  /hello:
    get:
      ..
```

Previously, Fabrikt ignored the server URL from the `openapi.yml`. Now it is passed to the client functions. The server URL can be customized for each call.

```kotlin
HelloClient.get(
  apiConfiguration = ApiConfiguration(basePath = "https://somewhere.com")
)
```

If the `openapi.yml` contains more than one server URL, the first one is used.

**Example for custom headers**

The reason for this is that I need to customize some headers for every single call, and it is not sufficient to add them to the `openapi.yml`.

```kotlin
HelloClient.get(
  apiConfiguration = ApiConfiguration(customHeaders = mapOf("SomeHeader" to "some value"))
)
```

If `customHeaders` contains a header that is already present in the request, it is replaced with the value from `customHeaders`.

### DIFF
--- a/src/test/resources/examples/ktorClient/client/ktor/KtorApiConfiguration.kt
+++ b/src/test/resources/examples/ktorClient/client/ktor/KtorApiConfiguration.kt
@@ -1,0 +1,24 @@
+package examples.ktorClient.client
+
+/**
+ * Configuration for the API.
+ * @property basePath The base URL path for the API
+ * @property customHeaders A map of custom HTTP headers to include in every request
+ */
+class ApiConfiguration(
+    val basePath: String = "https://api.example.com/v1",
+    val customHeaders: Map<String, String> = mapOf()
+) {
+    /**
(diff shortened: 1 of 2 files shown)


cjbooms:
LGTM

PR #557, opened 2026-03-04
Minor dependency cleanup

cjbooms:
Minor dependency cleanup


cjbooms:
@nickcaballero thanks for the PR earlier. I had a proper dig through and couldn't find much to improve on. I expected a bigger PR here, but just these small improvements was all I could see

nickcaballero:
Yeah, I took a quick look when you mentioned using BOMs and didn't find a whole lot either. Ktor was one that I thought could maybe use bundles, but the improvement was minor.

PR #562, opened 2026-03-13
fix: prevent phantom sealed interface for oneOf with shared discriminated parent

### LINKED ISSUE #561: SEALED_INTERFACES_FOR_ONE_OF generates non-existent interface name for oneOf property referencing discriminated type
## Description

When using `SEALED_INTERFACES_FOR_ONE_OF`, if a schema has a `oneOf` property that references subtypes of a discriminated parent type, Fabrikt 26.0.0+ generates code that implements a non-existent interface.

## Minimal Reproduction

```yaml
openapi: 3.0.3
info:
  title: Minimal Repro
  version: 1.0.0
paths:
  /actions:
    post:
      operationId: performAction
      requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/ActionRequest'
      responses:
        '200':
          description: OK
components:
  schemas:
    ParentAction:
      type: object
      discriminator:
        propertyName: actionType
        mapping:
          CHILD_A: '#/components/schemas/ChildActionA'
          CHILD_B: '#/components/schemas/ChildActionB'
      properties:
        actionType:
          type: string
          enum: [CHILD_A, CHILD_B]
      required: [actionType]

    ChildActionA:
      allOf:
        - $ref: '#/components/schemas/ParentAction'
        - type: object
          properties:
            fieldA:
              type: string

    ChildActionB:
      allOf:
        - $ref: '#/components/schemas/ParentAction'
        - type: object
          properties:
            fieldB:
              type: integer

    ActionRequest:
      type: object
      properties:
        id:
          type: string
        action:
          oneOf:
            - $ref: '#/components/schemas/ChildActionA'
            - $ref: '#/components/schemas/ChildActionB'
```

Running with `--http-model-opts 'SEALED_INTERFACES_FOR_ONE_OF'`, Fabrikt 26.0.0 generates:

```kotlin
data class ChildActionA(
  ...
) : ParentAction(), ActionRequestParentAction  // ActionRequestParentAction does not exist
```

With 25.9.2, the same spec correctly generates:

```kotlin
data class ChildActionA(
  ...
) : ParentAction()
```

## Expected behavior

When a `oneOf` property references subtypes of an already-discriminated parent, the subtypes should NOT also implement a sealed interface for that `oneOf` — the parent's discriminator already handles polymorphism.

## Versions affected

- **25.9.2**: Works correctly
- **26.0.0 – 26.2.0**: All produce the unresolved reference

The regression was introduced in commit `c85ff54` ("Fabrikt v26 - Breaking Changes (#533)").

## Workaround

Use Fabrikt 25.9.2 (latest working version).

thejeff77:
## Summary

- Fixes a bug where `isOneOfSuperInterface()` incorrectly returns `true` for `oneOf` schemas whose members all inherit from a common `allOf` super type with a discriminator, generating a phantom sealed interface (e.g. `ActionRequestParentAction`) that subtypes reference but is never emitted — causing compilation failures
- Adds a single condition: `!(isOneOfWhereAllTypesInheritFromACommonAllOfSuperType() && hasNoDiscriminator())` to `isOneOfSuperInterface()`, consistent with the existing pattern at `safeName()` (line 280)

## Repro

Given this spec structure:
```yaml
ParentAction:
  type: object
  discriminator:
    propertyName: actionType
    mapping:
      CHILD_A: '#/components/schemas/ChildActionA'
      CHILD_B: '#/components/schemas/ChildActionB'

ChildActionA:
  allOf:
    - $ref: '#/components/schemas/ParentAction'
    - type: object
      properties:
        fieldA:
          type: string

ChildActionB:
  allOf:
    - $ref: '#/components/schemas/ParentAction'
    - type: object
      properties:
        fieldB:
          type: integer

ActionRequest:
  type: object
  properties:
    action:
      oneOf:
        - $ref: '#/components/schemas/ChildActionA'
        - $ref: '#/components/schemas/ChildActionB'
```

**Before**: Fabrikt generates a phantom `ActionRequestParentAction` sealed interface. `ChildActionA` and `ChildActionB` implement it, but the interface class is never emitted → compilation error.

**After**: The inline `oneOf` correctly resolves to `ParentAction` (the existing discriminated parent), and no phantom sealed interface is generated.

## Test plan

- [x] All 202 existing unit tests pass
- [x] All end2end tests pass (ktor, okhttp, openfeign, spring-http-interface, models-jackson, models-kotlinx)
- [x] Verified fix against minimal repro spec — no phantom type generated
- [x] `polymorphicModels` test (top-level named `oneOf` with its own discriminator) still generates sealed interface correctly

Fixes #561

🤖 Generated with [Claude Code](https://claude.com/claude-code)


thejeff77:
Thanks for the review! A couple of responses:

**Re: double negatives** — agree it reads awkwardly. Happy to extract a named helper to make it clearer, e.g.:

```kotlin
private fun Schema.isRedundantOneOfForExistingDiscriminatedHierarchy(): Boolean =
    isOneOfWhereAllTypesInheritFromACommonAllOfSuperType() && hasNoDiscriminator()
```

Then the condition becomes simply `!isRedundantOneOfForExistingDiscriminatedHierarchy()`. Let me know if you'd like that refactor.

**Re: adding to an existing test spec** — I added it as a standalone test (`oneOfWithSharedDiscriminatedParent`) with a dedicated spec, and verified it fails without the fix and passes with it. Would you prefer I fold the schema into the existing `discriminatedOneOf` test spec instead? Happy to move it there if that's cleaner for the project's test organization.

thejeff77:
Updated per your review feedback:

1. **Extracted named helper** — `isRedundantOneOfForExistingDiscriminatedHierarchy()` eliminates the double negative. The condition now reads `!isRedundantOneOfForExistingDiscriminatedHierarchy()` with a KDoc explaining when a oneOf is redundant (all members share a common discriminated allOf parent, and the oneOf itself has no discriminator).

2. **Moved test into existing `discriminatedOneOf` spec** — added the regression case (ParentAction/ChildActionA/ChildActionB/ActionRequest) to the `discriminatedOneOf` test, including both Jackson and kotlinx expected outputs. Removed the standalone test fixture.

All tests pass (unit + end2end).

PR #574, opened 2026-04-08
fix: extend oneOf redundancy check to non-discriminated allOf hierarchies

### LINKED ISSUE #573: SEALED_INTERFACES_FOR_ONE_OF phantom interface for oneOf with non-discriminated allOf parent
## Description

Follow-up to #561. The fix in #562 handles the case where all `oneOf` members share a common `allOf` parent **with a discriminator**. But the same phantom sealed interface issue occurs when the parent has **no discriminator** (e.g., Kotlin sealed classes serialized without `@JsonTypeInfo`).

## Minimal Reproduction

```yaml
components:
  schemas:
    BaseError:
      type: object
      properties:
        message:
          type: string
        errorType:
          type: string
          enum: [VALIDATION, SERVER]
      required: [message, errorType]
    
    ValidationErr:
      allOf:
        - $ref: '#/components/schemas/BaseError'
    
    ServerErr:
      allOf:
        - $ref: '#/components/schemas/BaseError'
    
    ErrorWrapper:
      type: object
      properties:
        error:
          oneOf:
            - $ref: '#/components/schemas/ValidationErr'
            - $ref: '#/components/schemas/ServerErr'
```

Note: `BaseError` has **no discriminator** — unlike the `ParentAction` example from #561.

## Actual behavior (26.2.2)

```kotlin
data class ServerErr(...) : ErrorWrapperError  // ErrorWrapperError does not exist
```

Fabrikt generates the phantom `ErrorWrapperError` sealed interface reference but never emits the interface class → compilation failure.

## Expected behavior

```kotlin
data class ServerErr(...)  // no phantom interface
```

The `oneOf` should be recognized as redundant (all members share a common `allOf` parent), regardless of whether that parent has a discriminator.

## Root cause

`isOneOfWhereAllTypesInheritFromACommonAllOfSuperType()` in `KaizenParserExtensions.kt:332` requires the common parent to have a discriminator:

```kotlin
return if (maybeAllOfInFirstOneOf != null && maybeAllOfInFirstOneOf.hasDiscriminator()) {
    this.oneOfSchemas.all { it.allOfSchemas.contains(maybeAllOfInFirstOneOf) }
} else false
```

The `hasDiscriminator()` check is too strict. The `oneOf` is redundant whenever all members share a common parent — the parent doesn't need a discriminator for this to be true.

## Fix

PR #573 removes the `hasDiscriminator()` requirement from the common parent check. The `oneOf` itself still must have `hasNoDiscriminator()` (checked in `isRedundantOneOfForExistingHierarchy`) to preserve the behavior for intentional discriminated `oneOf` schemas.

thejeff77:
## Summary

Follow-up to #562. Extends the phantom sealed interface fix to handle `oneOf` schemas where all members inherit from a common `allOf` parent that has **no discriminator**.

### Root cause

`isOneOfWhereAllTypesInheritFromACommonAllOfSuperType()` required the common parent to have a discriminator (`hasDiscriminator()`). This meant non-discriminated parents (e.g., Kotlin sealed classes without `@JsonTypeInfo`) weren't recognized, and a phantom sealed interface was still generated.

### Fix

One-line change: remove the `hasDiscriminator()` requirement from the common parent check. The redundancy condition is: all `oneOf` members share a common `allOf` parent AND the `oneOf` itself has no discriminator. Whether the parent has a discriminator is irrelevant — if all members already inherit from the same type, the `oneOf` is redundant either way.

### Test

Extends the regression test from #562 with a non-discriminated variant:
- `BaseError` (no discriminator) ← `ServerErr`, `ValidationErr` via `allOf`
- `ErrorWrapper.error` has `oneOf: [ServerErr, ValidationErr]`
- Expected: `ErrorWrapper.error` resolves to `BaseError?`, no phantom `ErrorWrapperError` interface

Both Jackson and Kotlinx serialization expected outputs included.

### Results

- All model generator tests pass (both Jackson and Kotlinx)
- Full test suite: 36 failures (down from 39 on master — this fix resolves 3 pre-existing test failures)

Fixes #573

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
@@ -328,10 +328,11 @@ object KaizenParserExtensions {
 
     fun Schema.isOneOfWhereAllTypesInheritFromACommonAllOfSuperType(): Boolean {
         val maybeAllOfInFirstOneOf = this.oneOfSchemas?.firstOrNull()?.allOfSchemas?.firstOrNull()
-        // This identifies the OLD allOf-based polymorphism pattern where the BASE type has the discriminator
-        return if (maybeAllOfInFirstOneOf != null && maybeAllOfInFirstOneOf.hasDiscriminator())  {
-            this.oneOfSchemas.all { it.allOfSchemas.contains(maybeAllOfInFirstOneOf) }
-        } else false
+            ?: return false
+        // All oneOf members must share the same allOf parent — with or without a discriminator.
+        // When the parent has no discriminator (e.g., Kotlin sealed classes without @JsonTypeInfo),
+        // the oneOf is still redundant because the parent type already represents the hierarchy.
+        return this.oneOfSchemas.all { it.allOfSchemas.contains(maybeAllOfInFirstOneOf) }
     }
 
     fun Schema.isOneOfResolvingToAnyType(): Boolean {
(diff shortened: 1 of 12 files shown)


thejeff77:
@nickcaballero FYI

PR #580, opened 2026-04-20
Revert #574 and fix discriminator-less inline oneOf generation

cjbooms:
Reverts the functional change in #574 and replaces it with a more principled fix. While the previously generated code would compile, without the JsonType annotations, json serialization/deserialzation would have caused data loss if the sub-types had different properties.

### What #574 did: 

Removed the `hasDiscriminator()` guard from `isOneOfWhereAllTypesInheritFromACommonAllOfSuperType()`, which caused non-discriminated allOf hierarchies to suppress sealed interface generation entirely and resolve member types to the common parent (e.g. `BaseError?`).

Why this is wrong: `BaseError?` is not a valid deserialization target for a discriminator-less oneOf — Jackson cannot dispatch to the correct subtype, and `BaseError` is not a supertype of `ValidationErr/ServerErr` in the generated code anyway. The field must be `Any?` so Jackson captures the raw JSON as a `LinkedHashMap` without data loss.

### What this PR does instead:

 - Restores the `hasDiscriminator()` guard (revert)
 - Keeps the test schema added by #574, updated with distinct properties on each subtype to make it realistic
 - Fixes a related bug: when an inline oneOf property has no discriminator, the sealed interface (`ErrorWrapperError`) was being referenced on the subtypes but never emitted. ModelGenerator now generates the sealed interface from `PropertyInfo.Field` entries whose schema satisfies `isOneOfSuperInterface()`
 - Adds end-to-end Jackson tests proving that serialisation uses the runtime type (all fields present) and deserialisation is non-lossy via `LinkedHashMap`

Net generated output for a discriminator-less inline oneOf:

 `sealed interface ErrorWrapperError          // marker — compiles, type-safe to produce`
 `data class ValidationErr(...) : ErrorWrapperError`
 `data class ServerErr(...) : ErrorWrapperError`
 `data class ErrorWrapper(val error: Any? = null)  // Any? — honest, non-lossy on deserialisation`

### DIFF
--- a/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DiscriminatorlessOneOfTest.kt
+++ b/end2end-tests/models-jackson/src/test/kotlin/com/cjbooms/fabrikt/models/jackson/DiscriminatorlessOneOfTest.kt
@@ -1,0 +1,74 @@
+package com.cjbooms.fabrikt.models.jackson
+
+import com.cjbooms.fabrikt.models.jackson.Helpers.mapper
+import com.example.oneof.models.BaseErrorErrorType
+import com.example.oneof.models.ErrorWrapper
+import com.example.oneof.models.ServerErr
+import com.example.oneof.models.ValidationErr
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+/**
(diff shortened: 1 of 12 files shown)


cjbooms:
cc @thejeff77

PR #587, opened 2026-04-21
feat: opt-in Jackson DEDUCTION for discriminator-less inline oneOf

thejeff77:
## Summary

Follow-up to #580 (which reverted #574). Adds an opt-in path that recovers the typed-property ergonomics #574 was reaching for, without the data-loss problems that motivated the revert.

Per-schema vendor extension `x-fabrikt-jackson-deduction: true` on a discriminator-less inline `oneOf` property opts that property into typed polymorphism:

- The field is typed as the sealed super-interface (e.g. `DiagnosticReportFailure?`) instead of `Any?`.
- The interface gets `@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)` + `@JsonSubTypes(...)` so Jackson dispatches by inspecting which properties are present in the JSON.

Default behavior (#580: `Any?` + non-lossy `LinkedHashMap` fallback) is unchanged for any `oneOf` without the extension. The existing `DiscriminatorlessOneOfTest` still passes.

## Why per-schema extension (not a CLI flag)

DEDUCTION is inherently per-schema safe: it only works when subtypes have distinguishing required properties. A spec may have several discriminator-less `oneOf`s — only some of which qualify. A global flag would either force-enable on schemas that fail at runtime, or stay off for schemas that would benefit. The vendor extension lets the spec author opt in only where they've confirmed it's safe.

This also matches the precedent set by `x-jackson-include-non-null`, `x-json-merge-patch`, and `x-async-support` — all per-property/operation opt-ins for serialization-library-specific behavior.

## Generated output

For:

```yaml
DiagnosticReport:
  type: object
  properties:
    failure:
      x-fabrikt-jackson-deduction: true
      oneOf:
        - \$ref: '#/components/schemas/NetworkFailure'
        - \$ref: '#/components/schemas/DnsFailure'
```

Jackson:

```kotlin
data class DiagnosticReport(val failure: DiagnosticReportFailure? = null)

@JsonTypeInfo(use = JsonTypeInfo.Id.DEDUCTION)
@JsonSubTypes(
  JsonSubTypes.Type(value = NetworkFailure::class),
  JsonSubTypes.Type(value = DnsFailure::class),
)
sealed interface DiagnosticReportFailure
```

Kotlinx (extension ignored — no DEDUCTION equivalent):

```kotlin
data class DiagnosticReport(val failure: Any? = null)
@Serializable sealed interface DiagnosticReportFailure
```

## Constraint

Subtypes must have distinguishing required properties for deduction to succeed. The fixture (`NetworkFailure.retries` vs `DnsFailure.host`) demonstrates this; if subtypes have identical property sets, deserialization fails at runtime. Documented in the helper KDoc and on the test fixture YAML.

## Test plan

- [x] Unit: full `ModelGeneratorTest` + `KotlinSerializationModelGeneratorTest` suites pass — existing fixtures untouched, new fixtures added for the opt-in case (Jackson + kotlinx).
- [x] End-to-end: new `DiscriminatorlessOneOfWithDeductionTest` proves typed round-trip — `objectMapper.readValue(json, DiagnosticReport::class.java).failure` returns a real `NetworkFailure`/`DnsFailure` instance (not a `LinkedHashMap`), and `failure as NetworkFailure` succeeds.
- [x] No regression on `DiscriminatorlessOneOfTest` (the `Any?` baseline added in #580) — opt-in is purely additive.
- [x] Full `./gradlew test` (root + all end2end-tests modules) passes.

cc @cjbooms — this is the follow-up I mentioned would address the gap left after #580. Happy to iterate on naming (`x-fabrikt-jackson-deduction` vs `x-jackson-typeinfo-deduction` vs other) or on whether DEDUCTION should be the right mechanism at all. Marking draft so you can weigh in before we land it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ This library was built to take advantage of the complex modeling features availa
  - GraalVM Native Reflection Registration
  - Json Merge Patch (via `JsonNullable`) (add `x-json-merge-patch: true` to schemas)
  - Override Jackson Include NonNull (via `JsonInclude`) (add `x-jackson-include-non-null: true` to schemas)
- - Type discriminator-less inline `oneOf` properties as the sealed super-interface using Jackson's deduction-based polymorphism (add `x-fabrikt-jackson-deduction: true` to the `oneOf` schema; subtypes need distinguishing required fields)
+ - Type discriminator-less inline `oneOf` properties as the sealed super-interface using Jackson's deduction-based polymorphism (add `x-jackson-subtype-deduction: true` to the `oneOf` schema; subtypes need distinguishing required fields)
  
 as well as HTTP clients and controllers for a number of popular frameworks (see [Features](#features)).
 
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This library was built to take advantage of the complex modeling features availa
  - GraalVM Native Reflection Registration
  - Json Merge Patch (via `JsonNullable`) (add `x-json-merge-patch: true` to schemas)
  - Override Jackson Include NonNull (via `JsonInclude`) (add `x-jackson-include-non-null: true` to schemas)
+ - Type discriminator-less inline `oneOf` properties as the sealed super-interface using Jackson's deduction-based polymorphism (add `x-fabrikt-jackson-deduction: true` to the `oneOf` schema; subtypes need distinguishing required fields)
  
 as well as HTTP clients and controllers for a number of popular frameworks (see [Features](#features)).
 
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ This library was built to take advantage of the complex modeling features availa
  - GraalVM Native Reflection Registration
  - Json Merge Patch (via `JsonNullable`) (add `x-json-merge-patch: true` to schemas)
  - Override Jackson Include NonNull (via `JsonInclude`) (add `x-jackson-include-non-null: true` to schemas)
- - Type discriminator-less inline `oneOf` properties as the sealed super-interface using Jackson's deduction-based polymorphism (add `x-jackson-subtype-deduction: true` to the `oneOf` schema; subtypes need distinguishing required fields)
+ - Jackson's polymorphic subtype deduction (add `x-jackson-subtype-deduction: true` to the `oneOf` schema) to gain Sealed Interface polymorphism, without needing to declare a discriminator property. Requires subtypes to have distinguishing required fields.
  
 as well as HTTP clients and controllers for a number of popular frameworks (see [Features](#features)).
 
--- a/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinxSerializationAnnotations.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/model/KotlinxSerializationAnnotations.kt
@@ -101,6 +101,9 @@ object KotlinxSerializationAnnotations : SerializationAnnotations {
     override fun addPolymorphicSubTypesAnnotation(typeSpecBuilder: TypeSpec.Builder, mappings: Map<String, TypeName>) =
         typeSpecBuilder // not applicable
 
+    override fun addDeductionPolymorphicTypeAnnotation(typeSpecBuilder: TypeSpec.Builder, subTypes: List<TypeName>) =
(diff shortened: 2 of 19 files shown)


cjbooms:
This is a really nice feature I knew nothing about. Gives you polymorphism without requiring a discriminator property. 

Happy to accept this. Couple of things as you take it out of draft:
- requestsDeduction - I find this name jarring, can you rethink this Boolean flag. Perhaps something like: isSubTypeDeductionEnabled
- Do we need the fabrikt in the extension property, im not sure we do for others. Perhaps just x-jackson-subtype-deduction
- please limit the llm kdoc comments. 

Nice work, thank you

thejeff77:
Thanks @cjbooms! All three addressed in 9c8830c:

- **Extension renamed** to `x-jackson-subtype-deduction` (dropped the `fabrikt` prefix as suggested)
- **Helper renamed** `requestsJacksonDeduction()` → `requestsSubTypeDeduction()` to match
- **Boolean param renamed** `requestsDeduction` → `isSubTypeDeductionEnabled` on `oneOfSuperInterface()` and all five call sites
- **KDocs trimmed** on the new helper, the `SerializationAnnotations` method, the end-to-end test class, and the api.yaml fixture comment

Tests + end-to-end pass locally. Marking ready for review.

src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
```diff
@@ -376,6 +376,13 @@ object KaizenParserExtensions {
     fun Schema.isOneOfSuperInterfaceWithDiscriminator() =
         discriminator != null && discriminator.propertyName != null && isOneOfSuperInterface()
 
+    /** Per-schema opt-in for Jackson DEDUCTION-style polymorphism. Subtypes must have
+     *  distinguishing required fields or deserialization fails at runtime. */
+    fun Schema.requestsSubTypeDeduction(): Boolean =
```
cjbooms:
```suggestion
    fun Schema.isSubTypeDeductionEnabled(): Boolean =
```

Can we harmonise this function name terminology

README.md
```diff
@@ -32,6 +32,7 @@ This library was built to take advantage of the complex modeling features availa
  - GraalVM Native Reflection Registration
  - Json Merge Patch (via `JsonNullable`) (add `x-json-merge-patch: true` to schemas)
  - Override Jackson Include NonNull (via `JsonInclude`) (add `x-jackson-include-non-null: true` to schemas)
+ - Type discriminator-less inline `oneOf` properties as the sealed super-interface using Jackson's deduction-based polymorphism (add `x-jackson-subtype-deduction: true` to the `oneOf` schema; subtypes need distinguishing required fields)
```
cjbooms:
```suggestion
 - Jackson's polymorphic subtype deduction (add `x-jackson-subtype-deduction: true` to the `oneOf` schema) to gain Sealed Interface polymorphism, without needing to declare a discriminator property. Requires subtypes to have distinguishing required fields.
```

src/main/kotlin/com/cjbooms/fabrikt/model/KotlinxSerializationAnnotations.kt
```diff
@@ -101,6 +101,9 @@ object KotlinxSerializationAnnotations : SerializationAnnotations {
     override fun addPolymorphicSubTypesAnnotation(typeSpecBuilder: TypeSpec.Builder, mappings: Map<String, TypeName>) =
         typeSpecBuilder // not applicable
 
+    override fun addDeductionPolymorphicTypeAnnotation(typeSpecBuilder: TypeSpec.Builder, subTypes: List<TypeName>) =
```
cjbooms:
```suggestion
    override fun addPolymorphicSubTypeDeductionAnnotation(typeSpecBuilder: TypeSpec.Builder, subTypes: List<TypeName>) =
```

cjbooms:
Looks good. Just a few renames to harmonising terminology throughout the stack

cjbooms:
Hey @thejeff77 would you be willing to make these small suggestions so we can merge this?

cjbooms:
Thanks, great addition.

PR #591, opened 2026-04-26
Document how to run Fabrikt from the command line with JBang

wfouche:
If JBang is installed then it will download all the required JAR dependencies and start the Fabrikt JAR file with the specified parameters.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,18 @@ java -jar fabrikt.jar \
     --http-client-opts resilience4j
 ```
 
+Alternatively, using [JBang](https://www.jbang.dev/), the CLI can also be invoked as follows:
+
+```
+jbang run io.fabrikt:fabrikt:RELEASE \
+    --output-directory '/tmp' \
+    --base-package 'com.example' \
+    --api-file '/path-to-api/open-api.yaml' \
+    --targets 'client' \
+    --targets 'http_models' \
+    --http-client-opts resilience4j
+```
+
 ### Gradle w/ custom task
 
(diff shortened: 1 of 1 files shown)


cjbooms:
Thanks! 
Fabrikt tries to stay agnostic to specific tool usage, so we unfortunately decline this contribution. Please do not let this deter you from future contributions

wfouche:
> Fabrikt tries to stay agnostic to specific tool usag

@cjbooms - Docker, Maven, Gradle even Java are all tools. JBang is a tool to make it easier to consume Java artifacts.

I am just trying to make it easier for end-user to use Fabrikt.

Truely surprised at the decision that was taken. Please reconsider accepting this PR.

> jbang run io.fabrikt:fabrikt:RELEASE

Using the `RELEASE` tag ensure that JBang will automatically used the latest (stable) released version of Fabrikt. Now need for users to download the JAR file and run it with java -jar.

cjbooms:
Hi @wfouche, You make a fair point and clearly feel strongly about this, so how about this compromise:

To keep the README focused and avoid duplicating the long list of CLI arguments, I’d be open to including it as a concise one-liner note at the end of the Command Line section. What about something like below?

> **Tip**: You can also run the latest version without a manual download via [JBang](https://www.jbang.dev/):
> ```
> jbang run io.fabrikt:fabrikt:RELEASE [args...]
> ```

wfouche:
> jbang run io.fabrikt:fabrikt:RELEASE [args...]

Thanks! This is perfect. I'll update the PR accordingly.

PR #598, opened 2026-05-06
Escape reserved word query parameters in ktor clients

arthurljones:
### Background
When attempting to generate a `ktor` API client for [Immich](https://github.com/immich-app/), fabrikt generated invalid code for endpoints with query parameters that are kotlin reserved names (in this case, `for`). I dug in a bit and discovered that the Kotlin identifiers were being passed to KotlinPoet as literals (`%L`) rather than names (`%N`), so they weren't being escaped.

### Changes
I changed all four paths in `KtorClientGenerator`'s query param statement generation to use `%N` when referencing the Kotlin argument itself, and added a new `GET /reserved-invalid-words` endpoint to `examples/ktorClient/api.yaml` that covers those four paths.

### See Also
- [Immich OpenAPI spec usage of "for" as a param name](https://github.com/immich-app/immich/blob/dafe9d7966f0f5634a94e6ef9242400ce4646859/open-api/immich-openapi-specs.json#L6341)


ulrikandersen:
Awesome fix @arthurljones! Thanks for the contribution 🤗

PR #603, opened 2026-05-09
fix #602 handle requestBody required attribute correctly

### LINKED ISSUE #602: RequestBody "required" not handled correctly
### Description

According to the OpenAPI Specification: https://spec.openapis.org/oas/v3.0.4.html#fixed-fields-10 a RequestBody is not required by default, so it should be generated as optional (Kotlin nullable).
Only if `required: true` is provided the type can be non-nullable.

I can work on a PR.

### Minimal OpenAPI Spec

```yaml
/example:
    post:
      summary: Example Post with optional request body
      parameters:
        - name: param1
          in: header
          schema:
            type: string
      requestBody:
        description: optional body
        content:
          image/jpeg:
            schema:
              type: string
              format: byte
```

### Options Used

_No response_

### Fabrikt Version

27.0.2

pstorch:


cjbooms:
LGTM - thanks

Issue #309, opened 2024-08-23
Are 'Multipart Requests' supported?


b0m123:
Hi, I have a question whether Multipart requests are supported? (see https://swagger.io/docs/specification/describing-request-body/multipart-requests/) 
I'm trying to generate simple OkHttp client with model (here is my endpoint in openapi.yaml as example)
```
  /api/import
    post:
      operationId: importMedia
      summary: Import Media
      description: >
        TEST Import media file
      requestBody:
        required: true
        content:
          multipart/form-data:
            schema:
              type: object
              required:
                - interaction
              properties:
                metadata:
                  $ref: '#/components/schemas/MediaFileDto'
                mediaFile:
                  type: array
                  items:
                    type: string
                    format: binary
```
but I'm facing an error `Unresolved reference 'MultipartFormData'.`

To be clear I'm using [fabrikt-gradle-plugin](https://github.com/acanda/fabrikt-gradle-plugin) with this config
```
fabrikt {
	generate("Client") {
		apiFile = file("openapi.yaml")
		externalReferenceResolution = aggressive
		basePackage = "com.example"
		outputDirectory = file("${project.layout.buildDirectory.get().asFile}/generated")
		typeOverrides {
			datetime = OffsetDateTime
		}
		client {
			generate = enabled
			target = OkHttp
		}
		model {
			generate = enabled
			sealedInterfacesForOneOf = enabled
			ignoreUnknownProperties = enabled
			includeCompanionObject = enabled
		}
	}
}
```

Thanks in advance for any support

ulrikandersen:
Hi @b0m123,

I can confirm that requests with `multipart/form-data` does not currently lead to working code being generated. 

This applies to the `OkHttp` client as well as the controller interfaces for Micronaut, Spring and Ktor.

ulrikandersen:
One possible cause of action would be to add a special case to the [request statement builder](https://github.com/cjbooms/fabrikt/blob/master/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpSimpleClientGenerator.kt#L213) for operation.requestBody.getPrimaryContentMediaType()?.key == "multipart/form-data", which would construct the requestBody using MultipartBody.Builder().

mfhellh:
Any idea how to work around this in the case of Controller (Spring) interfaces?
Thanks in advance!

ulrikandersen:
[...]

PR #605, opened 2026-05-17
feature: multipart form data support for OkHttpClient and SpringController

pstorch:
Implements the feature request #309 for OkHttpClient and SpringController generators. 

Builds on top of the work from @ulrikandersen 
I've rebased his branch on current `master`, fixed merge conflicts and added support for Spring Controller.

Out of scope:
-  MicronautControllerInterfaceGenerator
-  KtorClient
-  KtorController
-  OpenFeign client
-  Spring client


pstorch:
I could verify the implementation of the SpringControllerInterfaceGenerator in my own SpringBoot Service.
But failed to verify the OkHttpClient Generator in my Android App. It's not working yet and it lacks some functionality. I need to be able to provide the filename and contentType. Still working on it.

I'm not sure if I can provide implementations for the other generators.

cjbooms:
Thanks for the efforts. We can merge in small increments if you prefer. One generator fixed is better than none

pstorch:
@cjbooms ok, I think it is ready for review. I tested OkHttpClient and SpringController in my own Projects as well.

ulrikandersen:
Hi @pstorch,

Thanks for picking this one up 💪 

A few comments from a quick read through:

1. Great to see e2e tests for OkHttp. They are really useful as they validate that the generated code actually compiles and work. **It would be nice if you can add e2e tests for Spring controller too.**
2. I don't think the Resilience4j service wrapper will compile for multipart endpoints. Meaning if someone generates with Resilience4j enabled and the spec has multipart endpoints the code will not compile. **Worth adding a test for this case too.** See `OkHttpEnhancedClientGenerator.kt` which calls `addIncomingParameters(parameters)` without the `multipartParameterToSpecBuilder` param.
3. `deriveMultipartParameters` (`ClientGeneratorUtils.kt`) is **never called**. Left over or should it be called? 🤔 
4. `isRequired` seems to be dropped on parameter-name-clash rename. In `GeneratorUtils.kt`, the MultipartParameter rebuilt with the `multipart_` prefix omits `isRequired`, so it silently defaults to false. **Pass isRequired = p.isRequired through.**

pstorch:
@cjbooms thanks for the review. I'll look into your suggestions.

pstorch:
@cjbooms adding a e2e test for the SpringController was a good idea. Found and fixed a bug.

I removed the unused function.

Still working on number 2 and 4.

pstorch:
@cjbooms I fixed the OkHttpEnhancedClientGenerator and the isRequired attribute for MultipartParameter. I also did a small refactoring to make the generated OkHttpClient more descriptive when working with MultipartParameters.

src/test/kotlin/com/cjbooms/fabrikt/generators/UpdateMultipartExpectedFiles.kt
```diff
@@ -0,0 +1,76 @@
...
+import com.cjbooms.fabrikt.model.Models
+import com.cjbooms.fabrikt.model.SourceApi
+import com.cjbooms.fabrikt.util.Linter
+import com.squareup.kotlinpoet.FileSpec
+import java.io.File
+import java.nio.file.Paths
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+class UpdateMultipartExpectedFiles {
```
cjbooms: Could we instead move to making use of 
[GeneratedCodeAsserter.kt](https://github.com/fabrikt-io/fabrikt/blob/master/src/test/kotlin/com/cjbooms/fabrikt/util/GeneratedCodeAsserter.kt) to assert, and update the examples inline with others?

src/test/kotlin/com/cjbooms/fabrikt/generators/MultipartTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    @BeforeEach
+    fun init() {
+        MutableSettings.updateSettings(
+            genTypes = setOf(CodeGenerationType.CLIENT),
+            clientTarget = ClientCodeGenTargetType.OK_HTTP,
+        )
+    }
+
+    @Test
+    fun `multipart form data generates correct client and model code`() {
```
cjbooms: I think this is already tested in `OkHttpClientGeneratorTest.kt`

  ` fun `correct api simple client is generated from a full API definition`(testCaseName: String)`

src/test/resources/examples/kdoc/api-special-chars.yaml
```diff
@@ -0,0 +1,14 @@
+openapi: 3.0.0
```
cjbooms: Is this new addtition used somewhere, I can't seem to find it?

cjbooms:
Great work, really high quality. Thanks for the contribution. Couple of small comments, mostly just around reusing existing test constructs

src/test/kotlin/com/cjbooms/fabrikt/generators/MultipartTest.kt
pstorch: Yes, as I included the "uploadMultpart" into the OkHttpClientGeneratorTest, this is obsolete

src/test/kotlin/com/cjbooms/fabrikt/generators/UpdateMultipartExpectedFiles.kt
pstorch: As I included the "uploadMultpart" into the OkHttpClientGeneratorTest, this is obsolete

src/test/resources/examples/kdoc/api-special-chars.yaml
pstorch: Looks like a leftover from the branch I based upon. Seems unused. I've removed it.

pstorch:
@cjbooms I cleaned up the PR and resolved your comments.

PR #609, opened 2026-06-03
Fix incorrect generation of `$ref` item schemas in inline parameter array schemas

fourls:
Fixes a problem introduced in #570, where for inline `type: array` schemas with a non-inline item schema, the item schema would be incorrectly marked as inline. This manifests as a schema name being synthesized for it rather than it using the ref'd name.


cjbooms:
Thanks @fourls

PR #610, opened 2026-06-03
Fallback to OAS key for property name in case of normalization conflict

### LINKED ISSUE #608: Model properties that differ only in case/capitalization are conflated (kotlinx.serialization, probably others)
### Description

I attempted to generate a ktor client and kotlinx.serialization models for the Ride With GPS API today (see https://ridewithgps.com/api/v1/openapi.yaml). This API apparently contains a model (`TrackPointTrip`) that includes two properties that differ only by capitalization (`t` and `T`). 

When processed through `fabrikt`, non-compiling code is generated in the model - `t` appears twice as a property name, and both instances have the annotation and comment for the `T` property from the spec (which appears last).

I am currently digging into the code for this to see if there's a straightforward option to resolve it. My current thought is to disambiguate names in collections of properties (after the normal deduplication by `oasKey`) by checking to see if the preferred `PropertyInfo.name` (i.e. `oasKey.camelCase()`) is present earlier in the list. If so, we could fall back to the raw `oasKey`, which will be unambiguous at this point.

This requires a coupling between bare properties and the collections/scopes they live in for the purposes of name disambiguation. At the moment I'm thinking through how to manage this coupling, any additional state it requires (likely on `PropertyInfo`), making sure that state stays in sync (ideally via a single source of truth), and that it doesn't invalidate assumptions elsewhere about the relationship between `oasKey` and `name` (hopefully this is fully encapsulated in `PropertyInfo` already).

I'm very open to chatting about how to handle this if it sounds like I'm not on the right track.

### Minimal OpenAPI Spec
```yaml
openapi: 3.0.0
components:
  schemas:
    NameDiffersOnlyInCase:
      type: object
      properties:
        x:
          type: string
          description: Description of lower case x
        X:
          type: string
          description: Description of capital X
```

### Code Produced
```kotlin
@Serializable
public data class NameDiffersOnlyInCase(
  /**
   * Description of capital X
   */
  @SerialName("X")
  public val x: String? = null,
  /**
   * Description of capital X
   */
  @SerialName("X")
  public val x: String? = null,
)
```

### Options Used

```--targets client --targets http_models --http-client-target KTOR --serialization-library KOTLINX_SERIALIZATION --instant-library KOTLIN_TIME_INSTANT --validation-library NO_VALIDATION --output-opts ADD_FILE_DISCLAIMER```

### Fabrikt Version

27.1.0

arthurljones:
Closes #608 

This PR resolves the above issue by adding some additional logic to `Schema.getInLinedProperties`: If two or more properties in an object have distinct raw OAS names that are the same after normalization, all properties in the conflicting group will fall back to using the raw OAS name instead of the normalized name in generated code.

I did dig into the usage of property names in discriminators, but that system is too complex for me to figure out in a few hours of reading and tinkering. However, it appeared to me that much of the discriminator code relies on the schema's property name matching `PropertyInfo.name`, which may only work for schema names that are already camel case. The `discriminatedOneOf` test case `api.yaml` only has camel case property names, so it doesn't seem to test that discriminator property names that get transformed during normalization work correctly. 

As an aside, this work did not fully fix my use case - I'm targeting JVM, so even though this change properly generates `val t` and `val T` in the model's constructor args, it doesn't compile due to JVM getter collisions (both have getters named `getT`). I'm ready to move on for the moment, though, so I figured I'd put up what I did accomplish for review.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
@@ -23,14 +23,16 @@ import com.reprezen.kaizen.oasparser.model3.OpenApi3
 import com.reprezen.kaizen.oasparser.model3.Schema
 
 sealed class PropertyInfo {
+    /**
+     * Property name used in generated code. For most properties, it is derived
+     * from a normalized version of [oasKey].
+     */
+    abstract val name: String
     abstract val oasKey: String
     abstract val typeInfo: KotlinTypeInfo
     abstract val schema: Schema
     open val isRequired: Boolean = false
     open val isInherited: Boolean = false
-
-    val name: String
-        get() = oasKey.camelCase()
 
     companion object {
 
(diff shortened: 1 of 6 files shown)


cjbooms:
Great work, thank you!

PR #613, opened 2026-06-04
fix #493: add CollectionFormat.CSV to @RequestLine for explode:false array query params

### LINKED ISSUE #493: List query parameters are exploded on the Feign client even when explode is false
## Problem

I have the following OpenAPI 3.1 schema:

```json
{
  ..., 
  "/v1/resource/": {
    "get": {
      "operationId": "getPage",
      "parameters": [
        {
          "in": "query",
          "name": "slug",
          "schema": {
            "type": "array",
            "items": {
              "type": "string"
            }
          },
       "explode": false,
       "style": "form"
      }
    ]
  }
}
```
According to the [OpenAPI docs](https://swagger.io/docs/specification/v3_0/describing-parameters/#query-parameters:~:text=form%20%E2%80%93%20/products%3Fcolor%3Dblue%2Cgreen%2Cred%20or%20/products%3Fcolor%3Dblue%26color%3Dgreen%2C%20depending%20on%20the%20explode%20keyword), `"style": "form"` with `"explode": false` should produce a comma separated list of values, e.g. `?slug=one,two,three`, but the generated Feign client ends up creating a request with exploded parameters.

Generated client method looks something like this:

```kotlin
@RequestLine("GET /v1/reagent/?slug={slug}")
@Headers("Accept: application/json")
public fun getResourcePage(
  @Param("slug") slug: List<String>? = null,
  @HeaderMap additionalHeaders: Map<String, String> = emptyMap(),
  @QueryMap additionalQueryParameters: Map<String, String> = emptyMap(),
): PaginatedResource
```

The suggested fix for Spring OpenFeign clients is to use the `@CollectionFormat(feign.CollectionFormat.CSV)` annotation to the `@FeignClient` class ([docs](https://docs.spring.io/spring-cloud-openfeign/docs/current/reference/html/#feign-collectionformat-support)), however, this seems  to only work for methods/classes annotated with `@RequestMapping` ([source](https://github.com/spring-cloud/spring-cloud-openfeign/pull/610/files#diff-4a0384ff3e4b36193d99940fe6fbcacf9d50cfa650bddf9c40c8fe064c9c0131R175)).

`@RequestLine` instead accepts a `collectionFormat` parameter on the annotation to provide this configuration ([source](https://github.com/OpenFeign/feign/blob/master/core/src/main/java/feign/RequestLine.java#L93)), however, each parameter in an OpenAPI spec can define `explode` independently, and feign doesn't seem to have the same `collectionFormat` configuration available on the `@Param` annotation. Instead, to configure the behavior on a per parameter basis, we'd probably have to utilize the `expand` parameter on the `@Param` annotation.

rvinzent:
When all array query parameters in a Feign operation have explode:false, add collectionFormat = CollectionFormat.CSV to @RequestLine so the client serializes them as comma-separated values (e.g. param=a,b,c) instead of the default exploded format (param=a&param=b&param=c).

If array params within the same operation have mixed explode values, a warning is logged and the default (exploded) behavior is used, since @RequestLine's collectionFormat applies to all collection params in the method.

Test fixture updated with a CSV-only endpoint (path 3) that gains the annotation, and a mixed-explode endpoint (path 4) that verifies the fallback with no collectionFormat.

Fixes #493

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/client/OpenFeignInterfaceGenerator.kt
@@ -16,6 +16,7 @@ import com.cjbooms.fabrikt.generators.client.ClientGeneratorUtils.getReturnType
 import com.cjbooms.fabrikt.generators.client.ClientGeneratorUtils.optionallyParameterizeWithResponseEntity
 import com.cjbooms.fabrikt.generators.client.ClientGeneratorUtils.simpleClientName
 import com.cjbooms.fabrikt.generators.client.metadata.OpenFeignAnnotations
+import com.cjbooms.fabrikt.generators.client.metadata.OpenFeignImports
 import com.cjbooms.fabrikt.model.ClientType
 import com.cjbooms.fabrikt.model.Clients
 import com.cjbooms.fabrikt.model.GeneratedFile
@@ -37,6 +38,7 @@ import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.asTypeName
 import com.squareup.kotlinpoet.buildCodeBlock
+import java.util.logging.Logger
 
 class OpenFeignInterfaceGenerator(
     private val packages: Packages,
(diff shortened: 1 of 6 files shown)


rvinzent:
Hey @cjbooms I finally got around to the alternative approach we discussed and I'd appreciate a review here! Thanks!

cjbooms:
Thanks @rvinzent

PR #614, opened 2026-06-08
fix #612 keep parameters with a default value non-nullable

### LINKED ISSUE #612: Header parameters with default value is kotlin nullable
### Description

I think as part of https://github.com/fabrikt-io/fabrikt/pull/603 headers with a default value are marked as nullable and for cases like in spring the param is then set per https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html#defaultValue()

### Minimal OpenAPI Spec

```yaml
/example:
  post:
    summary: Example Post with optional request body
    parameters:
      - name: param1
        in: header
        schema:
          type: string
          default: paramValue
```

### Options Used

_No response_

### Fabrikt Version

27.1.0

ulrikandersen:
## Description

Fixes #612.

PR #603 (#603) introduced nullability for non-required parameters in the generated Kotlin signatures. This had the side effect of marking parameters that declare a `default` value as nullable too.

A parameter with a default value is always populated by the framework — for example Spring sets it via [`@RequestParam(defaultValue = ...)`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html#defaultValue()) — so the generated type should remain non-nullable.

### Before

```kotlin
@RequestParam(value = "limit", required = false, defaultValue = "10")
limit: Int?,
```

### After

```kotlin
@RequestParam(value = "limit", required = false, defaultValue = "10")
limit: Int,
```

## Change

- Added `IncomingParameter.isNullable`, derived from `isRequired`.
- Overridden in `RequestParameter` so a present `defaultValue` keeps the type non-nullable.
- `toParameterSpecBuilder` now uses `isNullable` for both the standard and the `treatAnyTypeHeadersAsStrings` paths.
- Regenerated the affected controller/client example outputs.

This applies to all parameter locations (query, header, path) and restores the pre-#603 behaviour for parameters that carry a default value.

### DIFF
--- a/end2end-tests/spring/build.gradle.kts
+++ b/end2end-tests/spring/build.gradle.kts
@@ -1,0 +1,71 @@
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
+
+val generationDir = "$buildDir/generated"
+val apiFile = "$projectDir/openapi/api.yaml"
+
+sourceSets {
+    main { java.srcDirs("$generationDir/src/main/kotlin") }
+}
+
+plugins {
+    kotlin("jvm")
+}
+
+java {
+    sourceCompatibility = JavaVersion.VERSION_17
+    targetCompatibility = JavaVersion.VERSION_17
+}
+
+val springBootVersion = "3.4.3"
+
+dependencies {
+    implementation(platform(libs.jackson.bom))
+    implementation(libs.jakarta.validation.api)
+    implementation(libs.jackson.module.kotlin)
+    implementation(libs.jackson.databind)
+    implementation(libs.jackson.core)
+    implementation(libs.jackson.annotations)
+
+    // Spring Boot
+    implementation("org.springframework.boot:spring-boot-starter-web:$springBootVersion")
+
+    testRuntimeOnly(libs.junit.platform.launcher)
+    testImplementation(libs.bundles.junit)
+    testImplementation(libs.assertj.core)
+    testImplementation("org.springframework.boot:spring-boot-starter-test:$springBootVersion")
+}
+
+tasks {
+    val generateCode by creating(JavaExec::class) {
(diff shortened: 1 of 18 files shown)


y-marion:
Hi @ulrikandersen , is there anything that i can do to help get this ready? We are facing this "regression" in our generated interfaces at the moment and don't want to change everything.

ulrikandersen:
> Hi @ulrikandersen , is there anything that i can do to help get this ready? We are facing this "regression" in our generated interfaces at the moment and don't want to change everything.

I believe the fix is sound and ready, but would be great if you could take a look at the suggested change, and perhaps even verify it against your failing spec.

y-marion:
> > Hi @ulrikandersen , is there anything that i can do to help get this ready? We are facing this "regression" in our generated interfaces at the moment and don't want to change everything.
> 
> I believe the fix is sound and ready, but would be great if you could take a look at the suggested change, and perhaps even verify it against your failing spec.

I've applied your fix in the repos where we saw failures and it works as before pre-#603 👍.
I have no concerns regarding the changes you did; i think it's good that you added the new e2e spring test so that we cover more surface.

ulrikandersen:
> > > Hi @ulrikandersen , is there anything that i can do to help get this ready? We are facing this "regression" in our generated interfaces at the moment and don't want to change everything.
> > 
> > 
> > I believe the fix is sound and ready, but would be great if you could take a look at the suggested change, and perhaps even verify it against your failing spec.
> 
> I've applied your fix in the repos where we saw failures and it works as before pre-#603 👍. I have no concerns regarding the changes you did; i think it's good that you added the new e2e spring test so that we cover more surface.

Awesome. I'll merge and cut a release 🚀

Issue #366, opened 2025-02-13
Question about the client


EmilioLizama:
Hello! I recently started using this library and it's great, but I have some questions about the usage of the generated clients. Whenever I create the client, it creates a class for each path from the open api yaml file. It feels a bit tedious having to make an instance of every client for every path generated.  Is there a way to generate a single client, containing all of the path operations in a single file? Or

Also for the client generation I am using the OkHttp client.

cjbooms:
No, and I would estimate the work to make this change is many order of magnitudes greater than the instantiation problem. Open to PRs thought if you want to attempt the change though.

Djaler:
I think we should have an option to use tags for this, like in openapi-generator.

cjbooms:
See https://github.com/fabrikt-io/fabrikt/pull/575 which adds tag based grouping

drulmc:
@cjbooms I've just raised a PR which adds grouping by tags for the client libs that are generated: https://github.com/fabrikt-io/fabrikt/pull/615

The previous PR (#575) that was merged adds tag grouping but only targets the controllers, this one extends that further and applies it to the clients. I tried to open this issue again but can't, let me know if you want me to raise a new one 🙂.

PR #615, opened 2026-06-12
feat: group generated clients by tag

drulmc:
## Summary

This change adds support for grouping generated client classes by OpenAPI tag instead of URL path segment.

Today, client generation groups operations based on the URL structure, which can produce multiple generated classes for endpoints that belong to the same controller in the OpenAPI spec. That makes the generated client shape drift away from the way the spec is organised. With this change, clients can now be generated with a **1-to-1 mapping between controller/tag and client class**, so operations that share a tag are emitted into the same generated client.

To keep the change safe and backwards-compatible, this behaviour is **opt-in** via a new client codegen option: `GROUP_BY_TAG`. If the option is not provided, existing path-based grouping stays exactly as it is today.

Under the hood, the grouping decision is now centralised in shared client-generation logic and applied consistently across all supported client generators:
- OkHttp
- OpenFeign
- Spring HTTP Interface
- Ktor

The implementation also matches the existing controller generation semantics by grouping on the **first operation tag**, and it falls back to the existing path-based behaviour when tags are not present.

## Additional changes

- added snapshot test coverage for tag-grouped client generation
- added grouped fixtures for the `tagGrouping` example across supported client generators
- updated the README to document the new `GROUP_BY_TAG` option

## Other

Further addresses issue #366


cjbooms:
Nice one, thanks for the contribution

PR #619, opened 2026-06-21
fix simple types for multipart params in Spring Controller

pstorch:
follow up on #605


pstorch:
@cjbooms sorry, I found a bug in my last PR #605 it didn't handle other simpleTypes besides String correctly. That's why this fix.

cjbooms (approved):

cjbooms:
> @cjbooms sorry, I found a bug in my last PR #605 it didn't handle other simpleTypes besides String correctly. That's why this fix.

No worries, bugs in new features are expected. I'll cut a patch fix later today

PR #621, opened 2026-06-22
fix: don't crash code generation for a templated server url (#596)

### LINKED ISSUE #596: Support for Server Variables

Open API Spec defines a server variables section.
Using variables in the url section, causes the generation of the api spec to fail.

The example for server variables is copy-pasted directly from the openapi specification: https://spec.openapis.org/oas/latest.html#server-object-example

### Example Spec

```yaml
servers:
  - url: https://{username}.gigantic-server.com:{port}/{basePath}
    description: The production API server
    name: prod
    variables:
      username:
        # note! no enum here means it is an open value
        default: demo
        description: A user-specific subdomain. Use `demo` for a free sandbox environment.
      port:
        enum:
          - '8443'
          - '443'
        default: '8443'
      basePath:
        # open meaning there is the opportunity to use special base paths as assigned by the provider, default is "v2"
        default: v2
```

### Desired Output

```kotlin
I have no idea what to do with the servers object in the generated code, but I would recommend not crash the rest of the code generation, if nothing else is done with this section.
```

mvanhorn:
## Summary

Makes `OpenApi3.basePath()` resilient to a templated server URL so that an OpenAPI spec whose `servers[]` entry uses a templated URL (with a `variables` section) no longer crashes code generation. The `URI.create(...)` call is wrapped in `runCatching` and falls back to an empty base path when the URL cannot be parsed.

## Why this matters

A server object with a templated URL such as `https://{username}.gigantic-server.com:{port}/{basePath}` plus a matching `variables` block is valid OpenAPI, copied verbatim from the spec's own server-object example (issue #596).

The actual crash is in `OpenApi3.basePath()` (src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt:421), which is called during controller generation, for example from `AnnotationBasedControllerInterfaceGenerator` (src/main/kotlin/com/cjbooms/fabrikt/generators/controller/AnnotationBasedControllerInterfaceGenerator.kt:29). It does `URI.create(server.url)`, and the `{` and `}` characters in a templated URL are illegal in a URI, so `URI.create` throws `URISyntaxException`, which aborts the whole generation run.

I confirmed empirically that the `variables` section itself is parsed harmlessly by the kaizen parser. Removing only `variables` from the server objects does not stop the crash, because the templated `url` is the trigger. Hardening `basePath()` is the minimal change on the exercised codegen path: a templated server URL has no usable base path, so falling back to an empty string produces a controller with no request-mapping prefix rather than crashing. Normal server URLs are unaffected and still return their path.

## Testing

Adds `KaizenParserExtensionsTest` covering `basePath()`:
- a normal server URL returns its path (trailing slash trimmed)
- no servers returns an empty string
- a templated server URL does not throw and returns an empty string

Adds a `SpringControllerGeneratorTest` case that runs full controller generation against the issue's server-variables spec and asserts it generates files without crashing and emits no templated placeholder in the output.

Ran locally with JDK 21:
- `./gradlew :test --tests "com.cjbooms.fabrikt.util.KaizenParserExtensionsTest"` passes
- `./gradlew :test --tests "com.cjbooms.fabrikt.generators.SpringControllerGeneratorTest"` passes, plus the Micronaut and Ktor client generator suites and the `util` suite for regressions

Fixes #596

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/KaizenParserExtensions.kt
@@ -421,8 +421,9 @@ object KaizenParserExtensions {
     fun OpenApi3.basePath(): String =
         servers
             .firstOrNull()
-            ?.url?.let { URI.create(it) }
-            ?.path.orEmpty()
+            ?.url
+            ?.let { url -> runCatching { URI.create(url).path }.getOrNull() }
+            .orEmpty()
             .removeSuffix("/")
 
     fun OpenApi3.groupedPaths(groupingStrategy: GroupingStrategy): Map<String, Map<String, Path>> = when (groupingStrategy) {
--- a/src/test/kotlin/com/cjbooms/fabrikt/generators/SpringControllerGeneratorTest.kt
+++ b/src/test/kotlin/com/cjbooms/fabrikt/generators/SpringControllerGeneratorTest.kt
@@ -370,4 +370,46 @@ class SpringControllerGeneratorTest {
         }
     }
 
+    @Test
+    fun `generates controllers without crashing for a templated server url with variables`() {
+        // Issue #596: a server object with a templated url and a variables section caused
+        // URI.create() in basePath() to throw, aborting the whole generation run.
+        val spec =
+            """
+            openapi: 3.1.0
+            info:
+              title: test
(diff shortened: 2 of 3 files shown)


cjbooms:
LGTM - thanks for the contribution

Issue #374, opened 2025-02-20
Open enum support


nsprod:
Hey 👋 

I'm using an OpenAPI specification that makes use of `anyOf` to implement the "open enum pattern". 

This is what I'm doing:
```yaml
openapi: 3.0.0
info:
paths:
components:
  schemas:
    SomeObj:
      type: object
      properties:
        open_enum:
          $ref: '#/components/schemas/OpenEnum'
        close_enum:
          $ref: '#/components/schemas/CloseEnum'
    OpenEnum:
      anyOf:
        - type: string
          enum:
            - foo
            - bar
            - baz
        - type: string

    CloseEnum:
      type: string
      enum:
        - foo
        - bar
        - baz
```

Generated model:
```kotlin
@Serializable
public data class SomeObj(
  @SerialName("open_enum")
  public val openEnum: String? = null,
  @SerialName("close_enum")
  public val closeEnum: CloseEnum? = null,
)
```

When using kotlinx as serialization library the `OpenEnum` model is not generated and the property using this schema is of type `String`. Does fabrikt (+ kotlinx) have support for `anyOf`?

This pattern is very interesting to force the client (mobile app for eg) to handle unknown cases. It allows more flexibility for the server to add new cases without breaking the API contract or having to version each new addition to the enum

Example taken from the Swift OpenAPI generator documentation: https://swiftpackageindex.com/apple/swift-openapi-generator/1.7.0/documentation/swift-openapi-generator/useful-openapi-patterns#Open-enums-and-oneOfs

ulrikandersen:
While it is an interesting pattern I struggle to imagine how this would look in Kotlin's type system. 

In TypeScript it would be a union type (`type OpenEnum = "foo" | "bar" | "baz" | string;`).

Designing how this should look in Kotlin would be the first step to make this work.

cjbooms:
In Zalando we make use of `x-extensible-enum` for this pattern, and there is support in fabrikt for either treating them as enums for server-side, or just treating them as strings for client-side.

What has been a proposal before, is generating fallbacks for enums, which I think would be a better addition to fabrikt. See  proposed options: https://github.com/cjbooms/fabrikt/issues/256#issuecomment-1852292627

nsprod:
> While it is an interesting pattern I struggle to imagine how this would look in Kotlin's type system.
> 
> In TypeScript it would be a union type (`type OpenEnum = "foo" | "bar" | "baz" | string;`).
> 
[...]

PR #625, opened 2026-07-01
Add FAULT_TOLERANT_OPEN_ENUMS option

clementbeffa:
As discussed in #374, open enum support would be great for sharing swagger between iOS/Android. This option add support for that with minor changes reusing the FAULT_TOLERANT_ENUM generation.


cjbooms:
Apologies for the delay in reviewing this. Summer holidays :)

This LGTM. I had thought there was a lot more involved to the open enum pattern, that we had to preserve the unrecognized value. But this is a very nice pragmatic addition to support the modelling in OpenApi specs. 👏

PR #631, opened 2026-07-19
fix: make allOf polymorphism compile with kotlinx serialization

### LINKED ISSUE #627: kotlinx-serialization: missing type-override for discriminated-union child schemas
### Description

when using kotlinx_serialization, the discriminating property is not generated, it works with jackson.
Expected:

```kotlin
@Serializable
public data class ChildActionA(
  @SerialName("fieldA")
  public val fieldA: String? = null,
  @SerialName("actionType")
  override val actionType: ParentActionActionType = ParentActionActionType.CHILD_A
) : ParentAction(), ChildActionsAll

@Serializable
public data class ChildActionB(
  @SerialName("fieldB")
  public val fieldB: Int? = null,
  @SerialName("actionType")
  override val actionType: ParentActionActionType = ParentActionActionType.CHILD_B,
) : ParentAction(), ChildActionsAll
```

Actual:
```kotlin
@Serializable
public data class ChildActionA(
  @SerialName("fieldA")
  public val fieldA: String? = null,
) : ParentAction(), ChildActionsAll

@Serializable
public data class ChildActionB(
  @SerialName("fieldB")
  public val fieldB: Int? = null,
) : ParentAction(), ChildActionsAll
```

compile error:
```
Class 'ChildActionA' is not abstract and does not implement abstract base class member: val actionType: ParentActionActionType
Class 'ChildActionB' is not abstract and does not implement abstract base class member: val actionType: ParentActionActionType
```

example taken from: ([/src/test/resources/examples/discriminatedOneOf/api.yaml](https://github.com/fabrikt-io/fabrikt/blob/master/src/test/resources/examples/discriminatedOneOf/api.yaml))

### Minimal OpenAPI Spec

```yaml
openapi: 3.0.0
info:
paths:
components:
  schemas:
    ParentAction:
      type: object
      discriminator:
        propertyName: actionType
        mapping:
          CHILD_A: '#/components/schemas/ChildActionA'
          CHILD_B: '#/components/schemas/ChildActionB'
      properties:
        actionType:
          type: string
          enum:
            - CHILD_A
            - CHILD_B
      required:
        - actionType
    ChildActionA:
      allOf:
        - $ref: '#/components/schemas/ParentAction'
        - type: object
          properties:
            fieldA:
              type: string
    ChildActionB:
      allOf:
        - $ref: '#/components/schemas/ParentAction'
        - type: object
          properties:
            fieldB:
              type: integer
    ChildActionsAll:
      type: array
      default: []
      items:
        oneOf:
          - $ref: '#/components/schemas/ChildActionA'
          - $ref: '#/components/schemas/ChildActionB'
        discriminator:
          propertyName: actionType
          mapping:
            CHILD_A: '#/components/schemas/ChildActionA'
            CHILD_B: '#/components/schemas/ChildActionB'
```

### Options Used

https://try.fabrikt.io/?genTypes=HTTP_MODELS&serializationLibrary=KOTLINX_SERIALIZATION&modelOptions=SEALED_INTERFACES_FOR_ONE_OF&controllerTarget=SPRING&modelSuffix=&clientTarget=OK_HTTP&validationLibrary=NO_VALIDATION&externalRefResolutionMode=TARGETED

### Fabrikt Version

27.4.2

ulrikandersen:
With `KOTLINX_SERIALIZATION`, allOf child schemas of a discriminated parent generated code that did not compile: the sealed parent declared an abstract discriminator property that no child overrode. The parent also carried Jackson `@JsonTypeInfo`/`@JsonSubTypes` annotations regardless of the selected serialization library, and children were missing the class-level @SerialName mapping kotlinx needs to resolve subtypes.

kotlinx serialization emits the discriminator itself, and a property whose serial name matches the class discriminator fails at runtime, so the discriminator must not be a class property at all. The parent now gets `@JsonClassDiscriminator` (as the oneOf sealed interface path already does), children get `@SerialName` with their mapping key, and the abstract property and its overrides are omitted.

Inherited non-discriminator properties had a related defect: kotlinx includes superclass backing fields in the subclass serial state, so the generated open-property-with-constructor-passthrough pattern produced a duplicate serial name compile error. Super type properties are now generated as abstract for kotlinx and overridden in the subtypes.

Jackson output is unchanged.

Closes #627

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/generators/model/ModelGenerator.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/generators/model/ModelGenerator.kt
@@ -20,8 +20,6 @@ import com.cjbooms.fabrikt.generators.TypeFactory.createMutableMapOfMapsStringTo
 import com.cjbooms.fabrikt.generators.TypeFactory.createMutableMapOfStringToType
 import com.cjbooms.fabrikt.generators.TypeFactory.createSet
 import com.cjbooms.fabrikt.generators.ValidationAnnotations
-import com.cjbooms.fabrikt.generators.model.JacksonMetadata.basePolymorphicType
-import com.cjbooms.fabrikt.generators.model.JacksonMetadata.polymorphicSubTypes
 import com.cjbooms.fabrikt.model.SerializationAnnotations
 import com.cjbooms.fabrikt.model.Destinations.modelsPackage
 import com.cjbooms.fabrikt.model.GeneratedType
@@ -820,8 +818,8 @@ class ModelGenerator(
         constructorBuilder: FunSpec.Builder = FunSpec.constructorBuilder(),
     ): TypeSpec.Builder {
         this.addModifiers(KModifier.SEALED)
-            .addAnnotation(basePolymorphicType(discriminator.propertyName))
-            .modifiers.remove(KModifier.DATA)
+        serializationAnnotations.addBasePolymorphicTypeAnnotation(this, discriminator.propertyName)
+        this.modifiers.remove(KModifier.DATA)
 
         for (oneOfSuperInterface in oneOfSuperInterfaces) {
             val interfaceName = ModelNameRegistry.getBySchema(oneOfSuperInterface) ?: ModelNameRegistry.getOrRegister(oneOfSuperInterface)
@@ -848,8 +846,8 @@ class ModelGenerator(
         val maybeEnumDiscriminator = properties
             .firstOrNull { it.name == discriminator.propertyName }?.typeInfo as? KotlinTypeInfo.Enum
 
-        this.addAnnotation(polymorphicSubTypes(mappings, maybeEnumDiscriminator))
-            .addQuarkusReflectionAnnotation()
+        serializationAnnotations.addPolymorphicSubTypesAnnotation(this, mappings, maybeEnumDiscriminator)
+        this.addQuarkusReflectionAnnotation()
             .addMicronautIntrospectedAnnotation()
             .addMicronautReflectionAnnotation()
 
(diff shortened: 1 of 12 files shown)


src/main/kotlin/com/cjbooms/fabrikt/model/JacksonAnnotations.kt
```diff
@@ -42,8 +43,11 @@ object JacksonAnnotations : SerializationAnnotations {
     override fun addBasePolymorphicTypeAnnotation(typeSpecBuilder: TypeSpec.Builder, propertyName: String) =
```
cjbooms, line 43: Nit: the description says Jackson output is unchanged, but I think this is actually a small improvement. By passing `enumDiscriminator` through instead of `null`, enum discriminators now resolve to the enum constant name (e.g. `CHILD_A`) rather than the schema name. That looks like the right behavior — just worth calling out in the PR description so it does not read as accidental.

src/main/kotlin/com/cjbooms/fabrikt/generators/model/ModelGenerator.kt
```diff
@@ -899,6 +897,13 @@ class ModelGenerator(
             this.addSuperinterface(generatedType(packages.base, interfaceName))
         }
 
+        val superTypeDiscriminator = superType.schema.discriminator.takeIf { it.propertyName != null }
+            ?: superType.schema.getDiscriminatorForInlinedObjectUnderAllOf()
+        superTypeDiscriminator?.let { discriminator ->
+            val mappingKey = discriminator.mappingKeyForSchemaName(schemaName) ?: schemaName
```
cjbooms, line 903: Heads-up: the mapping key here comes from `mappingKeyForSchemaName`, which matches with `endsWith(schemaName)`. I tried to trip it up with `ChildActionA` and `XChildActionA` and both children ended up with `@SerialName("XCHILD")`, because `XChildActionA` ends with `ChildActionA`. The bug is pre-existing, but now that this PR adds `@SerialName` to allOf children it becomes user-visible. Might be worth a follow-up issue to tighten the match to the last path segment.

cjbooms (approved):
**Approve with non-blocking comments.**

Local verification passed:

- `./gradlew clean :test --tests KotlinSerializationModelGeneratorTest --tests ModelGeneratorTest` → BUILD SUCCESSFUL.
- `./gradlew :end2end-tests:models-kotlinx:build` → BUILD SUCCESSFUL (generated models compile and the 8 new E2E tests run).
- `gh pr checks 631` → `build pass 4m7s`.

The fix for #627 is correct: kotlinx allOf children now carry `@SerialName` and no longer redeclare the discriminator, so the generated code compiles.

src/main/kotlin/com/cjbooms/fabrikt/generators/model/ModelGenerator.kt
ulrikandersen: Good catch! Adressed in #650.

PR #640, opened 2026-08-10
Support remote api specification fetching

### LINKED ISSUE #521: FEATURE REQUEST: Support Remote Specs
Would it be possible to either have the `--api-file` parameter accept either a URL or a standard file path, or else create a new `--api-url` parameter that would allow for the ingestion of a remote spec that we are consuming? (If there's already a way to do that, I haven't been able to find it in any of the documentation or scanning through the code.)

Thanks!

mgmeiner:
--api-file and --api-fragment now accept a resolvable http:///https:// URL in addition to a local file path. When a URL is given, Fabrikt fetches the spec over HTTP at generation time instead of requiring it on disk.

Why

Teams that keep their OpenAPI spec in a central registry, artifact host, or another repo previously had to download it as a separate build step before running Fabrikt. This lets Fabrikt fetch it directly.

Compatibility & safety

- Backwards compatible: local-path invocations are untouched — same Paths.get/Files.notExists/readText calls, just reached through a branch. 
- Not a new trust boundary in the normal case — whoever controls --api-file already controls the whole build. Fetches are plain, unauthenticated GETs using the JVM's default ProxySelector (so standard -Dhttps.proxyHost/-Dhttp.nonProxyHosts settings are honored automatically); no TLS verification changes. README calls out that specs should only be fetched from trusted URLs.

closes #521

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,11 @@ java -jar fabrikt.jar \
     --http-client-opts resilience4j
 ```
 
+`--api-file` and `--api-fragment` also accept a resolvable `http://`/`https://` URL instead of a local path,
+in which case Fabrikt fetches the spec at generation time.
+Only fetch specs from URLs you trust — Fabrikt performs a plain, unauthenticated HTTP GET, so pointing it at an
+untrusted or attacker-influenced URL carries the same risk as fetching any other untrusted network resource.
+
 __Tip__: You can also run the latest version without a manual download via [JBang](https://www.jbang.dev/):
 
 ```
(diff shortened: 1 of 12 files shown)


ulrikandersen:
Looks good @mgmeiner. Nice addition 💯 

I would like @cjbooms's eyes on this too as he has [previously suggested](https://github.com/fabrikt-io/fabrikt/issues/521#issuecomment-3862545900) using a separate CLI argument. Using a single argument is simpler from the user's perspective, but could lead to unexpected behaviour. What do you think @cjbooms?

cjbooms:
Sorry for the delay. Yes, I think this is fine. Very happy to accept

cjbooms (approved):

mgmeiner:
Note: This change will not break the fabrikt-gradle-plugin. However, to make this feature available to projects using the Gradle plugin, the plugin configuration must be updated to accept flexible string paths for API file configurations.

Simply changing the property types from RegularFileProperty / ConfigurableFileCollection to CharSequence would probably break IDE path autocompletion. Maybe we should "overload" it at least at the gradle-plugin? 

fyi @acanda

PR #651, opened 2026-08-25
feat: implement OKHTTP_NON_NULL_RESPONSE_PAYLOADS client option

### LINKED ISSUE #649: ApiResponse.data is always nullable

In #120 / #124 the return type of client responses was changed to be non-null. However, `ApiResponse<T>` is always generated with `val data: T? = null`. This means even in the extremely common case where you declare a response body in the spec it is always required to do a null check on the caller side.

There is an open discussion in https://github.com/OAI/OpenAPI-Specification/issues/2236 about optional response payloads, where there's also this quote:

> [@jdegre](https://github.com/jdegre) - as soon as you keep content-type and schema for response body then it becomes mandatory to send that type of response.

Changing this as a default would be backwards incompatible, since calls which used to return an `ApiResponse(data=null)` would then start throwing `ApiException`. It could be solved by making this configurable with it possibly becoming the default if the spec issue is ever resolved.  

### Desired Output

```kotlin
data class ApiResponse<T>(val statusCode: Int, val headers: Headers, val data: T)
```

U1F984:
Closes #649

I decided to go with kotlinpoet everywhere (commit 2). There were some stale files in the repo, I've added verification so this doesn't happen again (commit 1). Commit 3 is the actual feature impl. 

Wrt. exceptions: I did add the `ApiSerializationException`, but it was weird implementing it behind the `NON_NULL` flag. I still think it would make sense to align with ktor, which also has an exception for network related issues, which is also a raw `IOException` today. If you agree, I can add it behind another flag or implement it without a flag (that would however impact backwards compatibility).


src/test/resources/examples/byteArrayStream/client/ApiModels.kt
U1F984, line 1: KotlinPoet adds `public` everywhere, adds explicit imports even for default imports and escapes identifiers, that's why there are so many changes.

src/main/kotlin/com/cjbooms/fabrikt/generators/client/OkHttpClientLibraryFiles.kt
```diff
@@ -0,0 +1,496 @@
...
+        AnnotationSpec.builder(Throws::class).addMember("%T::class", apiException(packages)).build()
+
+    fun apiModels(packages: Packages, nonNullDataPayloads: Boolean): FileSpec {
+        val dataType =
+            if (nonNullDataPayloads) TypeVariableName("T") else TypeVariableName("T").copy(nullable = true)
+        val dataParameter = ParameterSpec.builder("data", dataType)
+            .apply { if (!nonNullDataPayloads) defaultValue("null") }
+            .build()
+        return FileSpec.builder(packages.client, "ApiModels")
+            .indent("    ")
```
U1F984, line 47: This could be debated/changed. On the one hand, it makes the churn a little less on the files. But then again there's already quite some, and afaik the other generators use default indent.

src/test/resources/examples/okHttpClientNonNullResponsePayloads/client/HttpUtil.kt
```diff
@@ -0,0 +1,110 @@
...
+
+@Throws(ApiException::class)
+public fun <T> Request.execute(
+    client: OkHttpClient,
+    objectMapper: ObjectMapper,
+    typeRef: TypeReference<T>,
+): ApiResponse<T> = doRequest(client) { response ->
+    response.body?.string().isNotBlankOrNull()?.let {
+        objectMapper.readValue(it, typeRef)
+            ?: throw ApiException("[${response.code}]: Response body deserialized to null")
```
U1F984, line 60: These would've been ApiSerializationExceptions

Issue #656, opened 2026-08-26
RFC: A version-independent foundation for complete OpenAPI 3.1 and 3.2 support


mrhoeve:
### What would you like fabrikt to generate?

First of all, I want to emphasize that I am proposing this because I am genuinely impressed by Fabrikt and would like to contribute to its future.

Fabrikt produces clean, practical, and idiomatic Kotlin code, supports an unusually broad range of frameworks and serialization options, and places strong emphasis on generated-output stability. My recent Jackson 3 contribution gave me an even better appreciation of the existing codebase, its test suite, and the amount of work that has gone into making Fabrikt useful to a large Kotlin community.

I believe this is a foundation worth investing in.

## Objective

I would like to contribute towards complete and predictable support for OpenAPI 3.1 and 3.2, while designing the underlying architecture so that future OpenAPI versions can be added without another fundamental rewrite.

The intended guarantee would be:

- Every valid OpenAPI 3.x document can be loaded and analysed.
- Every described operation, request representation, response representation, webhook, callback, and security requirement remains available to the generation pipeline.
- Fabrikt generates strongly typed Kotlin wherever the specification can be represented safely.
- Constructs that cannot be expressed as a useful static Kotlin type use an explicit, lossless fallback such as a JSON value, raw body, or stream.
- Application-specific extensions and unknown schema vocabularies are preserved and reported, rather than silently discarded.
- Generated code always compiles for supported targets.
- Existing OpenAPI 3.0 output remains stable unless a separate, explicitly justified bug fix requires a change.

In this context, complete support does not mean pretending that every arbitrary custom extension can automatically become a meaningful Kotlin type. It means that every valid specification is processed safely and predictably, no operation or representation silently disappears, and users receive either strongly typed output or a clearly documented lossless fallback.

## Motivation

While investigating the remaining OpenAPI 3.1 gaps, it appears that many share the same underlying limitation: Fabrikt's generators depend directly on the OpenAPI 3.0-oriented Kaizen model.

The current 3.1 downgrading approach works well for targeted compatibility cases, but cannot preserve every construct introduced by JSON Schema 2020-12, OpenAPI 3.1, and OpenAPI 3.2. Examples include general type unions, boolean schemas, advanced reference semantics, multiple content representations with different schemas, webhooks, newer HTTP operations, and streaming media types.

Continuing to address these exclusively as isolated downgrade cases risks losing information before the generators have an opportunity to make an appropriate decision.

## Proposed architectural direction

[...]

PR #657, opened 2026-08-26
refactor: introduce a version-aware OpenAPI parsing boundary

mrhoeve:
## Summary

Introduce an internal, version-aware boundary for parsing OpenAPI documents while preserving the existing Kaizen-based generation pipeline and generated output.

This is the first incremental step discussed in #656. It creates a place where Fabrikt can retain information from newer OpenAPI specifications before compatibility transformations are applied.

## Changes

- Add an internal `ParsedOpenApiDocument` that retains the original source alongside the existing Kaizen model.
- Capture the OpenAPI version before applying compatibility transformations.
- Recognize OpenAPI 3.0, 3.1 and 3.2 version strings without changing generation behavior.
- Move direct `OpenApi3Parser` usage behind a dedicated Kaizen parser adapter.
- Route external OpenAPI documents loaded during model generation through the same adapter.
- Extract the shared YAML mapper and input-cleanup logic to avoid coupling and dependency cycles.
- Keep the existing `YamlUtils` API and downstream generator APIs unchanged.

## Motivation

Fabrikt currently applies compatibility transformations before passing documents to Kaizen. Some valid OpenAPI 3.1 information, such as multi-type declarations containing `null`, is consequently no longer available to later processing stages.

Preserving the source document and its version gives future changes access to that information while allowing the existing Kaizen model to remain in use during an incremental migration.

This PR deliberately does not change generated code or claim OpenAPI 3.1/3.2 generation support. Functional support can be added in separate, focused PRs once the required source model and processing stages are introduced.

## Testing

- Added coverage showing that an OpenAPI 3.1 nullable type array remains present in the retained source while the existing Kaizen compatibility representation is unchanged.
- Added version-detection coverage for OpenAPI 3.0, 3.1 and 3.2 documents.
- Added coverage for parsing external OpenAPI documents through the Kaizen adapter.
- Verified that the complete Gradle build passes.
- Verified that existing generated examples remain unchanged.

Related to #656.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
@@ -1,0 +1,36 @@
+package com.cjbooms.fabrikt.parser
+
+import com.cjbooms.fabrikt.util.OpenApi31Downgrader
+import com.cjbooms.fabrikt.util.YamlObjectMapper
+import com.reprezen.jsonoverlay.JsonLoader
+import com.reprezen.kaizen.oasparser.OpenApi3Parser
+import com.reprezen.kaizen.oasparser.model3.OpenApi3
+import java.net.URI
+import java.nio.file.Paths
+
+internal class ParsedOpenApiDocument(
+    val sourceContent: String,
+    val kaizenModel: OpenApi3,
+)
+
+internal object OpenApiDocumentParser {
+    fun parse(
+        input: String,
+        baseUri: URI = Paths.get("").toAbsolutePath().toUri(),
+        jsonLoader: JsonLoader? = null,
+    ): ParsedOpenApiDocument =
+        try {
+            val kaizenInput = YamlObjectMapper.instance.readTree(input)
+            OpenApi31Downgrader.downgradeIncompatibleElements(kaizenInput)
+            OpenApiInputCleaner.cleanEmptyTypes(kaizenInput)
+            val kaizenModel = OpenApi3Parser().parse(kaizenInput, baseUri.toURL(), false, jsonLoader) as OpenApi3
+            ParsedOpenApiDocument(input, kaizenModel)
+        } catch (ex: NullPointerException) {
+            throw IllegalArgumentException(
+                "The Kaizen openapi-parser library threw a NPE exception when parsing this API. " +
+                    "This is commonly due to an external schema reference that is unresolvable, " +
+                    "possibly due to a lack of internet connection",
+                ex,
+            )
+        }
+}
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
```diff
@@ -0,0 +1,37 @@
+package com.cjbooms.fabrikt.parser
+
+import com.cjbooms.fabrikt.util.OpenApi31Downgrader
+import com.cjbooms.fabrikt.util.YamlObjectMapper
+import com.reprezen.jsonoverlay.JsonLoader
+import com.reprezen.kaizen.oasparser.model3.OpenApi3
+import java.net.URI
+import java.nio.file.Paths
+
+internal class ParsedOpenApiDocument(
```
cjbooms: data class maybe?

src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
```diff
@@ -0,0 +1,37 @@
...
+    val sourceContent: String,
+    val version: OpenApiVersion?,
```
cjbooms, line 12: Should we assume a version if not present/parseable instead of null?

cjbooms (dismissed):
couple of nonblocking nit picks, but LGTM

src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
mrhoeve: I deliberately kept this nullable because a valid OpenAPI document is required to declare a parseable version. In this boundary, null therefore represents missing or unrecognized source information. Defaulting to 3.0 could hide malformed input and later cause us to apply the wrong version-specific semantics. I could rename it to declaredVersion to make that distinction clearer.

src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
mrhoeve: Agreed — `ParsedOpenApiDocument` is an immutable value holder, so a data class better expresses its role. Updated, thanks!

src/main/kotlin/com/cjbooms/fabrikt/parser/OpenApiDocumentParser.kt
cjbooms: fine as-is. As long as it is a conscious choice

cjbooms (approved):

PR #660, opened 2026-08-26
fix: bound duplicate model name suffixes

### LINKED ISSUE #622: Repeated query parameters with different enums generate ever‑growing …Extra class names
### Description

**Summary**
When two or more generated models resolve to the same name, Fabrikt disambiguates by appending the literal suffix **Extra**. The suffix is added once per collision, so N colliding models produce Foo, Foo**Extra**, Foo**ExtraExtra**, … (i.e. N‑1 copies of Extra). The name therefore grows linearly with the number of collisions, and once it crosses the filesystem's 255‑character filename limit the build fails with File name too long.
This is easy to hit with array query parameters that carry an inline enum and repeat across many endpoints — e.g. OData's $select / $orderby / $expand, which legitimately appear on every entity‑set endpoint, each with a different list of allowed values.

**Environment**
Noticed first in Fabrikt 27.1.0 (via ch.acanda.gradle.fabrikt 1.35.0)
Appears to be a regression — not observed on 27.0.0 (plugin 1.34.0)
But also also occurs in current Fabrikt Version 27.4.1 (ch.acanda.gradle.fabrikt:1.37.2)
OS: macOS / APFS (255‑char filename limit); the same would occur on ext4 (255 bytes)
Generation: model only

**Steps to reproduce**
Minimal spec — three endpoints, each with a $select array parameter whose enum differs:

```
{
  "openapi": "3.0.0",
  "info": {
    "title": "repro",
    "version": "1.0.0"
  },
  "paths": {
    "/a": {
      "get": {
        "operationId": "getA",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Alpha", "Beta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    },
    "/b": {
      "get": {
        "operationId": "getB",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Gamma", "Delta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    },
    "/c": {
      "get": {
        "operationId": "getC",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Epsilon", "Zeta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    }
  }
}
```

**Actual behavior**

Three enum classes are generated, named after the parameter and disambiguated with Extra:

```
Select.kt
SelectExtra.kt
SelectExtraExtra.kt
```

Scaling this to a realistic OData API (~50 such endpoints) keeps stacking the suffix until the file name exceeds the OS limit and generation aborts:

```
Select.kt
SelectExtra.kt
...
SelectExtraExtraExtra…Extra.kt        // ~49 × "Extra"
.../models/SelectExtraExtraExtra…Extra.kt: File name too long
```

**Expected behavior**

Name de‑duplication should use a bounded scheme — e.g. a numeric counter (Select, Select2, Select3, …) — so the generated name length stays roughly constant regardless of how many collisions occur and never exceeds the filesystem limit.

### Minimal OpenAPI Spec

```yaml
{
  "openapi": "3.0.0",
  "info": {
    "title": "repro",
    "version": "1.0.0"
  },
  "paths": {
    "/a": {
      "get": {
        "operationId": "getA",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Alpha", "Beta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    },
    "/b": {
      "get": {
        "operationId": "getB",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Gamma", "Delta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    },
    "/c": {
      "get": {
        "operationId": "getC",
        "parameters": [
          {
            "name": "$select",
            "in": "query",
            "schema": {
              "type": "array",
              "items": {
                "type": "string",
                "enum": ["Epsilon", "Zeta"]
              }
            }
          }
        ],
        "responses": {
          "200": {
            "description": "ok"
          }
        }
      }
    }
  }
}
```

### Options Used

_No response_

### Fabrikt Version

27.1.0 and above

mrhoeve:
## Summary

- Centralize unique model-name allocation in `ModelNameRegistry`.
- Preserve `Extra` for the first name collision and use numeric suffixes for subsequent collisions.
- Add model and Spring controller regression coverage based on the reproduction from #622.

## Motivation

Fabrikt currently resolves repeated model-name collisions by continually appending `Extra`, producing names such as:

```text
Select
SelectExtra
SelectExtraExtra
SelectExtraExtraExtra
```

For APIs containing many repeated inline enum query parameters, this can eventually exceed filesystem filename limits.

The new allocation sequence is bounded:

```text
Select
SelectExtra
SelectExtra2
SelectExtra3
```

Both regular schema registration and reference-based pre-registration now use the same allocation function.

## Compatibility

The existing base name and first collision remain unchanged as `Foo` and `FooExtra`. Numeric suffixes are introduced only for subsequent collisions, limiting generated-output changes to scenarios affected by the bug.

No unrelated golden files were changed.

## Testing

- Added a minimal specification with three `$select` array query parameters containing different inline enums.
- Added golden model coverage for `Select`, `SelectExtra`, and `SelectExtra2`.
- Added Spring controller coverage verifying that each parameter references the correct generated enum.
- Ran the targeted model and controller generator tests.
- Ran the complete clean build, including all end-to-end projects.

Development note: this implementation was prepared with assistance from OpenAI Codex and reviewed and validated by me.

Fixes #622

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/util/ModelNameRegistry.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/util/ModelNameRegistry.kt
@@ -30,14 +30,9 @@ object ModelNameRegistry {
         allocate: Boolean = true,
     ): String {
         val modelClassName = schema.toModelClassName(schemaInfoName, enclosingSchema, valueSuffix)
-        var suggestion = modelClassName
+        val suggestion = if (allocate) allocateUniqueName(modelClassName) else modelClassName
 
         if (allocate) {
-            while (allocatedNames.contains(suggestion)) {
-                suggestion += SUFFIX
-            }
-            allocatedNames.add(suggestion)
-
             val tag = resolveTag(schema, modelClassName)
             val replaced = tagToName.put(tag, suggestion)
             if (replaced != null) {
@@ -47,6 +42,18 @@ object ModelNameRegistry {
         }
 
         return suggestion
+    }
+
+    private fun allocateUniqueName(modelClassName: String): String {
+        if (allocatedNames.add(modelClassName)) return modelClassName
+
+        var collisionIndex = 1
+        while (true) {
+            val numericSuffix = if (collisionIndex == 1) "" else collisionIndex.toString()
+            val suggestion = "$modelClassName$SUFFIX$numericSuffix"
+            if (allocatedNames.add(suggestion)) return suggestion
+            collisionIndex++
+        }
     }
 
     private fun Schema.toModelClassName(
@@ -113,12 +120,7 @@ object ModelNameRegistry {
         val ref = Overlay.of(schema).jsonReference
         if (!referenceToName.containsKey(ref)) {
             val modelClassName = name.toModelClassName() + MutableSettings.modelSuffix
-            var suggestion = modelClassName
-            while (allocatedNames.contains(suggestion)) {
-                suggestion += SUFFIX
-            }
-            allocatedNames.add(suggestion)
-            referenceToName[ref] = suggestion
+            referenceToName[ref] = allocateUniqueName(modelClassName)
         }
     }
 
--- a/src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
+++ b/src/test/kotlin/com/cjbooms/fabrikt/generators/ModelGeneratorTest.kt
@@ -44,6 +44,7 @@ class ModelGeneratorTest {
             "arrays",
             "anyOfOneOfAllOf",
             "deepNestedSharingReferences",
+            "boundedModelNameCollisions",
             "defaultValues",
             "duplicatePropertyHandling",
             "enumExamples",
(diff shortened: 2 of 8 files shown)


ulrikandersen (approved):
Looks good to me 👍 The numeric suffix makes identifying the generated "extras" easier.

PR #663, opened 2026-08-28
Refactor OpenAPI parsing to preserve recursive source schemas

mrhoeve:
## Summary

Introduces a lossless, Fabrikt-owned source representation alongside the existing Kaizen model.

`SourceOpenApiDocument` retains the original input and unmodified parsed document. Kaizen compatibility transformations are now applied to a deep copy, preventing information from the original OpenAPI document from being lost before future processing can use it.

The initial `SourceSchema` model represents:

- object and boolean schemas;
- single and multiple schema types;
- equivalent nullable semantics across OpenAPI 3.0, 3.1, and 3.2;
- `$ref` values and `$ref` siblings;
- unknown and future keywords through the original schema node;
- recursively nested `properties`, `items`, `allOf`, `anyOf`, `oneOf`, `not`, and `additionalProperties`;
- exact JSON Pointer locations, including escaped property and component names.

## Motivation

Fabrikt currently applies compatibility transformations before passing a document to Kaizen. Although this allows the existing generators to keep working, OpenAPI 3.1 and 3.2 information may be simplified or discarded in the process.

This change establishes a lossless parsing boundary and the first version of a parser-independent schema model. It allows future OpenAPI 3.1 and 3.2 work to use the original schema semantics while keeping the existing Kaizen-based generation pipeline unchanged.

This is intended as an incremental architectural step related to #656, not as a claim of complete OpenAPI 3.1 or 3.2 generation support.

## Compatibility

Existing generators continue to use the Kaizen model. This PR does not change generated code or existing golden files.

## Testing

Added contract tests covering:

- source preservation before Kaizen compatibility transformations;
- OpenAPI version detection for 3.0, 3.1, and 3.2;
- equivalent nullable representations across these versions;
- boolean schemas;
- `$ref` siblings and unknown keywords;
- recursively nested schema structures;
- JSON Pointer escaping.

The full build passes successfully, including all end-to-end projects:

```text
BUILD SUCCESSFUL
126 actionable tasks: 126 executed
```

Refs #656

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
@@ -1,0 +1,38 @@
+package com.cjbooms.fabrikt.parser
+
+import com.fasterxml.jackson.databind.JsonNode
+
+internal sealed interface SourceSchema {
+    val location: String
+    val node: JsonNode
+}
+
+internal data class SourceBooleanSchema(
+    override val location: String,
+    override val node: JsonNode,
+    val allowsAnyValue: Boolean,
+) : SourceSchema
+
+internal data class SourceObjectSchema(
+    override val location: String,
+    override val node: JsonNode,
+    val types: Set<SourceSchemaType>,
+    val reference: String?,
+) : SourceSchema
+
+internal enum class SourceSchemaType(
+    val value: String,
+) {
+    ARRAY("array"),
+    BOOLEAN("boolean"),
+    INTEGER("integer"),
+    NULL("null"),
+    NUMBER("number"),
+    OBJECT("object"),
+    STRING("string"),
+    ;
+
+    companion object {
+        fun from(value: String): SourceSchemaType? = entries.firstOrNull { it.value == value }
+    }
+}
--- a/src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
@@ -18,6 +18,13 @@ internal data class SourceObjectSchema(
     override val node: JsonNode,
     val types: Set<SourceSchemaType>,
     val reference: String?,
+    val properties: Map<String, SourceSchema>,
+    val items: SourceSchema?,
+    val allOf: List<SourceSchema>,
+    val anyOf: List<SourceSchema>,
+    val oneOf: List<SourceSchema>,
+    val not: SourceSchema?,
+    val additionalProperties: SourceSchema?,
 ) : SourceSchema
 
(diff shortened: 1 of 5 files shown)


src/test/kotlin/com/cjbooms/fabrikt/parser/SourceOpenApiDocumentParserTest.kt
```diff
@@ -0,0 +1,172 @@
...
+                """
+                Value:
+                  type: [string, 'null']
+                """,
+            ),
+        )
+
+    private fun document(
+        version: String,
+        schemas: String,
```
cjbooms: NIt: Suggest you rename this to `componentSchemas` and add named arguments in the call functions. It took me awhile to figure out what was going on with these tests, and I think that will help the next reader.
Also perhaps rename the function to `createOpenApi3Document()`

src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
```diff
@@ -0,0 +1,45 @@
...
+    BOOLEAN("boolean"),
+    INTEGER("integer"),
+    NULL("null"),
+    NUMBER("number"),
+    OBJECT("object"),
+    STRING("string"),
+    ;
+
+    companion object {
+        fun from(value: String): SourceSchemaType? = entries.firstOrNull { it.value == value }
```
cjbooms: This is no longer lossless. Perhaps you want a better model to capture Unrecognised values?
Fine if you accept this, but let's discuss

cjbooms (changes requested):

src/test/kotlin/com/cjbooms/fabrikt/parser/SourceOpenApiDocumentParserTest.kt
mrhoeve: Thanks, agreed. I’ve renamed the helper to `createOpenApi3Document()`, renamed the parameter to `componentSchemas`, and added named arguments to all call sites. This makes the intent of the test setup much clearer.

src/main/kotlin/com/cjbooms/fabrikt/parser/SourceSchema.kt
mrhoeve: Good catch — I agree that dropping unknown type values contradicts the goal of a lossless source model.

I’ve changed `SourceSchemaType` into a sealed interface with `Recognised` values and an explicit `Unrecognised(value)` representation. The parser now preserves every textual type value, including unknown future values. I’ve also added tests for both a single unrecognised type and a mixed array containing recognised and unrecognised types.

Does this match the kind of model you had in mind?

cjbooms (approved):

PR #665, opened 2026-08-29
feat: add official Maven plugin for code generation (#664)

### LINKED ISSUE #664: Add an official Fabrikt Maven plugin
## Motivation

Fabrikt already provides good Gradle integration. The README documents both a custom `JavaExec` task and the separately maintained Fabrikt Gradle plugin, which offers a convenient DSL for configuring code generation.

The current Maven integration is more cumbersome. The [[Maven section in the README](https://github.com/fabrikt-io/fabrikt#maven)](https://github.com/fabrikt-io/fabrikt#maven) recommends using `exec-maven-plugin`. This requires Maven users to manually configure Fabrikt's main class, classpath, lifecycle phase, CLI arguments and generated source directories.

I would like to propose adding an official Maven plugin that provides Maven users with an integration comparable to the existing Gradle experience.

The plugin would remain a wrapper around the existing Fabrikt CLI. It would not require a new internal Fabrikt API.

## Proposed artifact and versioning

The Fabrikt repository would publish an additional artifact:

```text
io.fabrikt:fabrikt-maven-plugin
```

The Maven plugin and Fabrikt would always have the same version:

```text
io.fabrikt:fabrikt:X.Y.Z
io.fabrikt:fabrikt-maven-plugin:X.Y.Z
```

Version `X.Y.Z` of the Maven plugin would therefore always execute version `X.Y.Z` of Fabrikt.

Users would only configure the plugin version. A separate `fabriktVersion` property and a compatibility matrix would not be necessary.

The plugin could be added to the existing repository as an additional build artifact and released through the same release process as Fabrikt.

## Implementation approach

For each configured Maven execution, the plugin would:

1. Locate the Fabrikt executable JAR with the same version as the plugin.
2. Start Fabrikt in an isolated JVM.
3. Convert the Maven configuration into Fabrikt CLI arguments.
4. Execute Fabrikt for the configured OpenAPI specification.
5. Fail the Maven build if Fabrikt returns a failure.
6. Register the generated Kotlin source directories with the Maven project.

Using a separate JVM prevents Fabrikt's dependencies from conflicting with Maven's plugin classloader.

The existing Fabrikt CLI remains the source of truth for argument parsing and validation.

## One execution per OpenAPI specification

Every OpenAPI specification would be configured as a separate Maven execution.

This makes the input file, base package, lifecycle binding and Maven logging explicit for each API.

```xml
<plugin>
    <groupId>io.fabrikt</groupId>
    <artifactId>fabrikt-maven-plugin</artifactId>
    <version>${fabrikt.version}</version>

    <configuration>
        <arguments>
            <serializationLibrary>JACKSON_3</serializationLibrary>
            <validationLibrary>JAKARTA</validationLibrary>
        </arguments>
    </configuration>

    <executions>
        <execution>
            <id>generate-customer-api</id>
            <phase>generate-sources</phase>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputFile>src/main/openapi/customer.yaml</inputFile>
                <arguments>
                    <basePackage>com.example.customer</basePackage>
                </arguments>
            </configuration>
        </execution>

        <execution>
            <id>generate-order-api</id>
            <phase>generate-sources</phase>
            <goals>
                <goal>generate</goal>
            </goals>
            <configuration>
                <inputFile>src/main/openapi/order.yaml</inputFile>
                <arguments>
                    <basePackage>com.example.order</basePackage>
                </arguments>
            </configuration>
        </execution>
    </executions>
</plugin>
```

Every execution must have an effective `inputFile` and `basePackage`, either configured directly or inherited from the general configuration where appropriate.

## General configuration and execution overrides

Configuration on the plugin itself provides shared defaults.

Execution-specific configuration overrides only the values that differ for that API.

The precedence would be:

```text
Fabrikt CLI defaults
    < general plugin configuration
    < execution-specific configuration
```

In the example above, both executions inherit:

```text
serializationLibrary = JACKSON_3
validationLibrary = JAKARTA
```

Each execution supplies its own `inputFile` and `basePackage`.

An execution can override a shared setting:

```xml
<execution>
    <id>generate-legacy-api</id>
    <phase>generate-sources</phase>
    <goals>
        <goal>generate</goal>
    </goals>
    <configuration>
        <inputFile>src/main/openapi/legacy.yaml</inputFile>
        <arguments>
            <basePackage>com.example.legacy</basePackage>
            <serializationLibrary>JACKSON</serializationLibrary>
        </arguments>
    </configuration>
</execution>
```

All other shared settings remain inherited.

For repeatable arguments, an execution-specific list would replace the shared list for that argument.

## CLI argument mapping

Entries inside `<arguments>` would map from camelCase XML names to Fabrikt's kebab-case CLI options.

For example:

```xml
<arguments>
    <basePackage>com.example.customer</basePackage>
    <serializationLibrary>JACKSON_3</serializationLibrary>
</arguments>
```

Would be translated to:

```text
--base-package com.example.customer
--serialization-library JACKSON_3
```

This mapping would be generic. The Maven plugin would not need a dedicated Java or Kotlin property for every Fabrikt CLI argument.

Repeatable options could be expressed with nested values:

```xml
<arguments>
    <targets>
        <value>http_models</value>
        <value>client</value>
    </targets>
</arguments>
```

And translated to:

```text
--targets http_models
--targets client
```

Boolean and repeatable arguments would follow the syntax supported by the existing Fabrikt CLI.

This keeps the Maven plugin small and allows it to forward new Fabrikt CLI options without duplicating Fabrikt's argument definitions.

## Lifecycle behaviour

The plugin would follow normal Maven lifecycle behaviour.

Users who want generation to run automatically during a normal build bind each execution to `generate-sources`:

```xml
<phase>generate-sources</phase>
```

Those executions then run as part of:

```shell
mvn compile
mvn package
mvn install
mvn deploy
```

Users who only want manual generation can leave out the phase and invoke a named execution explicitly:

```shell
mvn fabrikt:generate@generate-customer-api
```

Multiple named executions can be invoked using Maven's normal goal syntax:

```shell
mvn fabrikt:generate@generate-customer-api \
    fabrikt:generate@generate-order-api
```

The plugin would not introduce custom lifecycle behaviour or a non-standard aggregate goal.

## Default output directory

The default Fabrikt output directory would be:

```text
${project.build.directory}/generated-sources
```

This produces the following source structure:

```text
target/generated-sources/src/main/kotlin
target/generated-sources/src/test/kotlin
```

Generated files then follow their normal package structure:

```text
target/generated-sources/src/main/kotlin/com/example/customer/Customer.kt
target/generated-sources/src/main/kotlin/com/example/order/Order.kt
```

The plugin registers the main Kotlin directory as a Maven source root and, when present, the test Kotlin directory as a Maven test source root.

All executions use the same generated source tree. Their required base packages provide the logical separation.

An execution may override the output directory when necessary.

## Cleaning and existing generated files

The Fabrikt Maven plugin would only generate and place code. It would not delete existing generated files or maintain a cache of previous runs.

In particular, it would not:

- Delete the output directory.
- Delete files produced by previous executions.
- Store timestamps or input hashes.
- Maintain generated-file manifests.
- Attempt to determine whether generation can be skipped.

Users decide whether they want a clean generation by using Maven's standard `clean` lifecycle:

```shell
mvn clean compile
```

Or:

```shell
mvn clean fabrikt:generate@generate-customer-api
```

Running without `clean` may overwrite files with the same path, while files that are no longer generated may remain present. This should be documented clearly.

Keeping the default output below `target` ensures that Maven's normal `clean` goal removes it.

## Scope

This proposal would not:

- Replace or change the existing Gradle integration.
- Replace or change the Fabrikt CLI.
- Introduce a public internal Fabrikt API.
- Duplicate Fabrikt's CLI validation.
- Maintain a separate Fabrikt/plugin compatibility matrix.
- Add custom incremental-build or cleanup behaviour.
- Add non-standard Maven lifecycle semantics.

The plugin is intended to be a small, predictable Maven adapter around the existing executable Fabrikt CLI.

## Testing

An initial implementation should include integration tests covering at least:

- A single OpenAPI specification.
- Multiple executions in one Maven project.
- Shared configuration inherited by executions.
- Execution-specific overrides.
- CamelCase-to-kebab-case argument conversion.
- Repeatable CLI arguments.
- Registration of generated main and test source roots.
- Compilation of the generated Kotlin sources.
- Propagation of Fabrikt failures to the Maven build.
- Explicit invocation using `goal@execution-id`.
- Lifecycle-bound generation during `generate-sources`.
- Behaviour with and without `mvn clean`.

## Questions

Before starting the implementation, I would appreciate feedback on the following:

1. Would an official Maven plugin be a welcome addition to Fabrikt?
2. Is maintaining it as an additional artifact in this repository acceptable?
3. Is `io.fabrikt:fabrikt-maven-plugin` the preferred artifact name?
4. Does lockstep versioning between Fabrikt and the Maven plugin fit the current release model?
5. Does the proposed configuration structure align with the project's preferred style?

If this direction is acceptable, I would be happy to work on an initial implementation and its integration tests.

mrhoeve:
## Summary

Adds an official Fabrikt Maven plugin as proposed in #664.

The plugin is published as `io.fabrikt:fabrikt-maven-plugin` and uses the same version as the Fabrikt executable it invokes.

## Implementation

- Invokes the existing Fabrikt executable JAR in a separate JVM.
- Keeps the Fabrikt CLI as the source of truth for argument parsing and validation.
- Maps camelCase Maven configuration elements to Fabrikt's kebab-case CLI arguments.
- Supports shared plugin configuration and execution-specific overrides.
- Supports repeatable arguments using nested `<value>` elements.
- Supports both lifecycle-bound and explicit goal execution.
- Uses `${project.build.directory}/generated-sources` as the default output directory.
- Registers generated main and test Kotlin source directories with Maven.
- Leaves cleaning and incremental behaviour to Maven's standard lifecycle.
- Publishes the plugin and Fabrikt executable with matching versions.
- Documents the new Maven integration in the README.

The implementation does not introduce a new internal Fabrikt API or change existing code-generation behaviour.

## Testing

Added unit and integration tests covering:

- CLI argument conversion.
- Boolean and repeatable arguments.
- Shared configuration and execution-specific overrides.
- Multiple OpenAPI specifications in one Maven project.
- Lifecycle-bound generation.
- Explicit `goal@execution-id` invocation.
- Generated source registration.
- Compilation of generated Kotlin sources.
- Fabrikt failure propagation.
- Behaviour with and without `mvn clean`.
- Plugin and Fabrikt version alignment.

The integration tests execute real Maven consumer projects against the locally built plugin.

`./gradlew clean build` passes successfully, including all existing Fabrikt, playground and end-to-end tests.

Closes #664


cjbooms (changes requested):
Very high quality submission again. Thanks. But definitely let's move it to its own repo. We can keep version in lockstep using Github Actions instead of co-locating.

PR #676, opened 2026-09-08
Introduce generator-facing schema contracts

mrhoeve:
## Summary

This PR introduces the first cohesive foundation for the native OpenAPI generation path discussed in #656.

It:

- models normalized, version-aware source schema semantics;
- introduces generator-facing schema contracts that do not depend on Kaizen types;
- adapts the existing Kaizen-backed schemas to those shared contracts;
- adds native schema type classification for OpenAPI 3.0, 3.1, and 3.2 semantics;
- introduces internal selection between legacy and native schema documents.

The existing Kaizen-backed path remains the default. The native mode is deliberately internal at this stage and is not exposed as a public CLI option yet.

## Motivation

The current generators consume Kaizen-specific schema types directly, which makes incremental support for modern OpenAPI and JSON Schema semantics difficult.

The new contracts establish a boundary between parsing and generation. This allows the existing implementation to continue through the legacy adapter while later PRs can build a Fabrikt-owned native representation behind the same generator-facing interface.

This is the first of the grouped follow-up PRs described in #673. It is intentionally limited to the schema boundary and classification foundation; model, endpoint, server, and client generation remain separate follow-up changes.

## Compatibility

- Existing generated output remains unchanged.
- `LEGACY` remains the effective default.
- No new public CLI option is introduced.
- No existing golden files are modified.

## Testing

- Added focused tests for normalized schema semantics.
- Added contract and legacy-adapter tests.
- Added native type-classification tests.
- Added OpenAPI 3.0, 3.1, and 3.2 schema-document selection coverage.
- The full project build, including end-to-end projects and generated-output checks, is used for final verification.

Part of #656.

### DIFF
--- a/src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
+++ b/src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
@@ -1,0 +1,127 @@
+package com.cjbooms.fabrikt.parser
+
+import com.cjbooms.fabrikt.model.OasType
+
+internal sealed interface GeneratorSchemaTypeClassification {
+    data class Resolved(
+        val type: OasType,
+        val nullable: Boolean,
+    ) : GeneratorSchemaTypeClassification
+
+    data class Unsupported(
+        val reason: Reason,
+    ) : GeneratorSchemaTypeClassification
+
+    enum class Reason {
+        NEVER_SCHEMA,
+        MULTIPLE_NON_NULL_TYPES,
+        INCONSISTENT_COMPOSITION_TYPES,
+    }
+}
+
+internal object GeneratorSchemaTypeClassifier {
+    fun classify(schema: GeneratorSchema): GeneratorSchemaTypeClassification =
+        when (schema) {
+            is GeneratorBooleanSchema ->
+                if (schema.allowsAnyValue) {
+                    GeneratorSchemaTypeClassification.Resolved(OasType.Any, false)
+                } else {
+                    GeneratorSchemaTypeClassification.Unsupported(GeneratorSchemaTypeClassification.Reason.NEVER_SCHEMA)
+                }
+            is GeneratorObjectSchema -> classifyObjectSchema(schema)
+            else -> error("Unknown generator schema implementation: ${schema::class.qualifiedName}")
+        }
+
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
```diff
@@ -0,0 +1,127 @@
...
+        val type: OasType,
+        val nullable: Boolean,
+    ) : GeneratorSchemaTypeClassification
+
+    data class Unsupported(
+        val reason: Reason,
+    ) : GeneratorSchemaTypeClassification
+
+    enum class Reason {
+        NEVER_SCHEMA,
```
cjbooms, line 16: I'm not really sure how never schema's work. My read was it's a way to explicitly reject certain proprties, such as `isAdmin: true` being sent.
What does Unsupported mean here, does it mean fabrikt gen will fail hard, or just skip this schema?

cjbooms (approved):

src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
mrhoeve: Thanks, that is a good question. Your interpretation of a `false` schema is correct: it describes a schema for which no value can ever be valid. When used for a property, it can therefore prohibit that property from being present. This differs from `{ const: false }`, which permits the property but only with the value `false`.

`Unsupported` in this PR is only an intermediate classification result; by itself it neither skips the schema nor makes generation fail. However, your comment makes clear that grouping `NEVER_SCHEMA` under that general name is ambiguous and could lead to incorrect downstream behaviour.

I will address this in the first native model-generation follow-up by explicitly distinguishing between:

- a semantically uninhabitable schema, for which no valid instance can exist;
- a valid schema that Fabrikt cannot represent yet;
- a schema that can safely use a more general Kotlin fallback type.

The eventual behaviour will then depend on where the schema occurs. An optional property with a `false` schema can be omitted, a `false` branch in `oneOf` or `anyOf` can be eliminated, while a required property, request body, or an `allOf` containing `false` makes the surrounding contract impossible to satisfy and should produce a clear generation error rather than silently falling back to `Any`.

I will add focused tests for these contexts as part of that follow-up. This does not affect the existing legacy pipeline, but it is important to resolve before exposing the native pipeline publicly.

Thanks for pointing this out, and for reviewing and merging the foundational contracts!

PR #677, opened 2026-09-13
Handle false schemas through shared generator contracts

mrhoeve:
## Summary

This PR reframes the original native model-generation work around the strangler approach discussed in the review.

It keeps the existing `ModelGenerator` as the only generation path. `GeneratorSchemaDocument` continues to expose Kaizen-adapted component schemas by default, while also making native schema semantics available by source location for functionality that Kaizen cannot represent.

The first migrated capability is OpenAPI 3.1 boolean `false` schemas:

- optional properties defined by an uninhabitable schema are omitted;
- required properties degrade to nullable `Any`;
- generated required fallback properties include KDoc explaining why the fallback was necessary;
- references and compositions resolving to `false` are classified consistently.

The previous `LEGACY`/`NATIVE` mode selection and parallel native generator have been removed. Further parser migration can now happen incrementally behind the shared contracts without introducing a second generator or a user-visible parser choice.

## Compatibility

Existing generation continues through the established Kaizen-backed behavior unless a native semantic is explicitly consumed through the shared document.

Existing golden output remains unchanged.

## Testing

- Added focused coverage for the hybrid generator document.
- Added classification tests for direct, referenced and composed uninhabitable schemas.
- Added model-generation coverage for optional and required `false`-schema properties.
- Ran the complete build, including the existing golden fixtures and end-to-end projects.

```text
./gradlew clean build
BUILD SUCCESSFUL
```

Relates to #656.


src/main/kotlin/com/cjbooms/fabrikt/cli/CodeGenerator.kt
```diff
@@ -14,22 +14,34 @@ import com.cjbooms.fabrikt.generators.controller.KtorControllerInterfaceGenerato
...
 import com.squareup.kotlinpoet.FileSpec
 import java.nio.file.Path
 
-class CodeGenerator(
+class CodeGenerator internal constructor(
     private val packages: Packages,
     private val sourceApi: SourceApi,
     private val srcPath: Path,
     private val resourcesPath: Path,
+    private val schemaGenerationMode: SchemaGenerationMode,
```
cjbooms: Can we just default this instead of having the additional constructor?

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -49,6 +54,14 @@ sealed class PropertyInfo {
             api: OpenApi3,
             enclosingSchema: Schema? = null,
             additionalRequiredFields: Collection<String> = emptySet(),
+        ): Collection<PropertyInfo> = topLevelProperties(settings, api, enclosingSchema, additionalRequiredFields, null)
+
+        internal fun Schema.topLevelProperties(
```
cjbooms, line 47: could we avoid the second signature and just default schemaDocument to null in the original?

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -251,6 +284,23 @@ sealed class PropertyInfo {
                                 }
                         }
                     }.filterNotNull()
+                    .plus(
```
cjbooms: Is this duplicate code needed? Looks like this should already be included from above

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -261,6 +311,12 @@ sealed class PropertyInfo {
                 mainProperties
             }
         }
+
+        private fun GeneratorSchemaDocument?.isUninhabitable(schema: Schema): Boolean =
```
cjbooms: Can this be moved into GeneratorSchemaDocument, or other, to reduce the duplicate delcarations of this classify-as-uninhabitable logic

cjbooms (changes requested):

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -1,5 +1,9 @@
 package com.cjbooms.fabrikt.model
 
+import com.cjbooms.fabrikt.parser.GeneratorBooleanSchema
+import com.cjbooms.fabrikt.parser.GeneratorSchemaDocument
+import com.cjbooms.fabrikt.parser.GeneratorSchemaTypeClassification
+import com.cjbooms.fabrikt.parser.GeneratorSchemaTypeClassifier
```
cjbooms: On closer inspection, is this not the wrong direction? Should they not be hidden behind the new OpenApi3Document abstraction?

cjbooms: I have added a section to Agents.md to try and avoid future abstraction leaks

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
mrhoeve: Done — I’ve collapsed this into the original topLevelProperties function and defaulted schemaDocument to null, so there is only one signature.

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
mrhoeve: You’re right. I verified that Kaizen retains the property entry for a boolean false schema, so the existing property traversal already sees it. I’ve removed the duplicate block and now handle the uninhabitable classification within that traversal.

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
mrhoeve: Done — the classification is now centralized in GeneratorSchemaDocument.isUninhabitableAt(...), and both property handling and schema validation delegate to that method.

src/main/kotlin/com/cjbooms/fabrikt/model/PropertyInfo.kt
```diff
@@ -1,5 +1,6 @@
 package com.cjbooms.fabrikt.model
 
+import com.cjbooms.fabrikt.parser.GeneratorSchemaDocument
```
cjbooms: This is still leaking in. GeneratorSchemaDocument should instead move into  OpenApiSchema

mrhoeve: (I was still working on it ;-))

Resolved — `PropertyInfo` no longer imports, accepts, or retains `GeneratorSchemaDocument` (or any other parser-internal schema type). The parser now binds the document-scoped classification as a neutral query when constructing `OpenApi3Document`, and generation code only asks that facade whether a schema is uninhabitable. I kept the query on `OpenApi3Document` because resolution and classification are document-scoped and reference-aware; no parser representation crosses the boundary.

src/main/kotlin/com/cjbooms/fabrikt/model/OpenApiModel.kt
```diff
@@ -90,19 +92,22 @@ class OpenApiSchema(
     override fun toString(): String = kaizen.toString()
 }
 
-class OpenApi3Document(
+class OpenApi3Document internal constructor(
     internal val kaizen: OpenApi3,
+    private val isUninhabitableAt: (String) -> Boolean,
 ) {
+    constructor(kaizen: OpenApi3) : this(kaizen, { false })
```
cjbooms: Same comment as before. Do we need separate constructors?

src/main/kotlin/com/cjbooms/fabrikt/model/OpenApiModel.kt
```diff
@@ -23,22 +23,23 @@ import com.reprezen.kaizen.oasparser.model3.Server
  */
 class OpenApiSchema(
     internal val kaizen: Schema,
+    private val isUninhabitableAt: (String) -> Boolean = { false },
```
cjbooms, line 26: this works for now, but I wonder should we be injecting the entire parallel schema model from our new parser

src/main/kotlin/com/cjbooms/fabrikt/model/OpenApiModel.kt
cjbooms: Confirmed load-bearing: `RelativeSchemaHandler.kt`, `SchemaParserExtensionsTest.kt`, and `OpenApiModelTest.kt` all call the single-arg form from outside `model/`, only reachable via this secondary constructor since the primary is `internal`. Fix: drop `internal` from the primary, default the lambda there (matches `OpenApiSchema`'s own pattern one class up), delete the secondary — same call sites keep compiling. One caveat: that makes the lambda param public to every caller, not just the parser; `OpenApiDocumentParser.kt:19` is currently the only two-arg caller, so nothing else would misuse it today, but it's worth deciding if that's intentional or if the two-arg form should stay parser-only.

src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
```diff
@@ -13,26 +15,58 @@ internal sealed interface GeneratorSchemaTypeClassification {
...
-                    GeneratorSchemaTypeClassification.Unsupported(GeneratorSchemaTypeClassification.Reason.NEVER_SCHEMA)
-                }
-            is GeneratorObjectSchema -> classifyObjectSchema(schema)
-            else -> error("Unknown generator schema implementation: ${schema::class.qualifiedName}")
+    fun classify(
+        schema: GeneratorSchema,
+        resolve: (GeneratorSchema) -> GeneratorSchema = { it },
+    ): GeneratorSchemaTypeClassification = classify(schema, resolve, mutableSetOf())
+
+    private fun classify(
```
cjbooms: `resolve` and `visiting` thread through every private function here (`classify`, `classifyObjectSchema`, `inferType`, `hasInconsistentCompositionTypes`) purely as plumbing for the one real caller (`GeneratorSchemaDocument.isUninhabitable`); every other caller uses the identity default. Suggest making this a `class` instead of an `object`, constructed with `resolve` bound and `visiting` as private instance state — the 5 private functions drop both params, and the companion `classify(schema, resolve)` entry point keeps today's call-site syntax unchanged (just constructs an instance internally).

src/main/kotlin/com/cjbooms/fabrikt/model/OpenApiModel.kt
mrhoeve: Thanks for confirming the call sites. I’ve applied exactly that simplification: the primary constructor is public, the callback defaults to false there, and the secondary constructor is removed. The existing single-argument call sites remain unchanged.

I’m intentionally accepting the public callback for now: it matches the OpenApiSchema pattern and exposes no parser-internal type. Restricting the two-argument form would require another construction path; if the callback is later replaced by a facade-owned native backing model, we can revisit that visibility then.

src/main/kotlin/com/cjbooms/fabrikt/model/OpenApiModel.kt
mrhoeve: I think a facade-owned representation of the complete parallel schema may well be the right eventual backing for OpenApiSchema. I kept this PR to the narrow semantic query because injecting GeneratorSchema directly would cross the parser/generator boundary codified in #679.

The callback lets the parser provide the document- and reference-aware answer without exposing its internal model. As the native parser replaces more Kaizen behaviour, we can evolve the facade’s internal backing as a separate step without changing generation code.

src/main/kotlin/com/cjbooms/fabrikt/parser/GeneratorSchemaTypeClassifier.kt
mrhoeve: Done — GeneratorSchemaTypeClassifier is now an instance with the resolver bound in the constructor and visiting kept as private state. The companion classify(schema, resolve) entry point preserves the existing call-site syntax while removing the recursive plumbing from the private functions.

cjbooms (approved):

PR #683, opened 2026-09-14
Add option to keep base url dynamic for ktor clients

F43nd1r:
Since https://github.com/fabrikt-io/fabrikt/pull/516, the `defaultRequest.url` of Ktor is no longer used, since the `ApiConfiguration` hardcodes an url. 
With multiple and/or dynamic server configuration, we now need to carry around an `ApiConfiguration` to every call. 

This problem can even be seen in the [samples](https://github.com/fabrikt-io/fabrikt/blob/master/end2end-tests/ktor-client-jackson/src/test/kotlin/com/cjbooms/fabrikt/clients/KtorClientJacksonTest.kt#L61), where the wiremock url is provided to both 
```kotlin
        defaultRequest {
            url(wiremock.baseUrl())
        }
``` 
and 
```kotlin
private fun apiConfig() = ApiConfiguration(basePath = wiremock.baseUrl())
```
which then is passed into every call via 
```kotlin
apiConfiguration = apiConfig()
```

---

This PR adds `ClientCodeGenOptionType.DYNAMIC_BASE_URL`, which prevents hardcoding of the url in `ApiConfiguration`, which means we no longer need to provide ApiConfiguration in every call when the hardcoded url might not match.


cjbooms:
Seems reasonable. Thanks for the contribution

cjbooms (approved):

PR #684, opened 2026-09-14
Add option to strip operation id prefix up to a separator

### LINKED ISSUE #682: Remove redundant information from client method name

I'd like to have an option to remove prefixes from the operation IDs:

Some OpenAPI producers use globally unique operation IDs with a controller/resource prefix:

```yaml
paths:
  /api/events:
    get:
      operationId: Events_GetEvents
```
Fabrikt currently generates:
```kotlin
class ApiEventsClient {
  suspend fun eventsGetEvents(...)
}
```
When operations are already separated into generated client classes, the prefix is redundant. I'd like to see:
```kotlin
class ApiEventsClient {
  suspend fun getEvents(...)
}
```

Possible solutions for transforming method names:
- a boolean configuration which strips the operation id before non-alphanumeric symbols
- a configuration which strips the operation id before a configurable separator
- a configuration with some sort of regex-replace for operation ids
- a extension property like x-fabrikt-method-name which takes precedence over the operation id.

### Example Spec

```yaml
openapi: 3.0.0
info:
  title: Example API
  version: 1.0.0
paths:
  /api/events:
    get:
      operationId: Events_GetEvents
      responses:
        "200":
          description: OK
  /api/v2/events:
    get:
      operationId: V2Events_GetEvents
      responses:
        "200":
          description: OK
```

### Desired Output

```kotlin
class ApiEventsClient {
  suspend fun getEvents(...)
}
class ApiV2EventsClient {
  suspend fun getEvents(...)
}
```

F43nd1r:
Closes #682 by implementing one of the suggested options.

OpenAPI spec generators often use prefixes to make sure operation ids are unique across the spec. However, when generating multiple api clients, we already have the separation and only need uniqueness per api client.
This PR allows us to strip these unnecessary prefixes.

Example:
```yaml
openapi: "3.0.0"
info:
  title: Test API
  version: "1.0"
paths:
  /events:
    get:
      operationId: Events_V2_GetEvents
      responses:
        '200':
          description: Success
```
generates 
```kotlin
class ApiEventsV2Client {
  suspend fun eventsV2GetEvents(...)
}
```
This PR now generates with  `clientOperationIdSeparator = "_"`:
```kotlin
class ApiEventsV2Client {
  suspend fun getEvents(...)
}
```

This is mostly cosmetic but helps when names are already long to keep an overview.


cjbooms:
Thanks for the contribution. I wonder though if this is more generalisable? Rather than performing the substring just for client generation, should it be done for all operations so controllers interfaces benefit also?

F43nd1r:
@cjbooms for sure, there are more general solutions - e.g. I mentioned regex-replace in [the issue](https://github.com/fabrikt-io/fabrikt/issues/682). 
This one I was able to implement in one evening so I just did that, but we can take a step back and discuss what the solution should be 🙂  Did you have one in mind?

cjbooms:
Closed in favour of #698 

Thanks @F43nd1r I have added a more generic solution using regex as you described

Issue #639, opened 2026-08-06
Selectively override types based on vendor extension fields


patrickstocklin:
### What would you like fabrikt to generate?

Hello,

I have been having a good time using the CLI for generating Kotlin clients for my services - thank you for this project. I ran into a scenario where the client options have been selected to override all `date-time` format fields to `LocalDateTime` to move away from the default `OffsetDateTime` types, but we also require support of `Instant` types as well now across the same endpoints / models.

Since the current CLI version provides options to override these types across the board, I am curious how we'd possibly enhance the generator to support selective overrides if it sounds agreeable to the maintainers.

My first thought was to decorate the spec's target fields with vendor extension values that the generator would respect and override, but it would have to play nicely with the pre-existing options to override all types such as `DATETIME_AS_INSTANT`. I am curious if this is something that others may have an appetite for or if this is an idea that would not be possible/supported by the project. 

Thank you.

### Example Spec

```yaml
...
  properties:
    createdDateTime:
      type: string
      format: date-time
      x-kotlin-type: java.time.LocalDateTime
    createdInstant:
      type: string
      format: date-time
      x-kotlin-type: java.time.Instant
...
```

### Desired Output

```kotlin
...
@param:JsonProperty("createdDateTime")
@get:JsonProperty("createdDateTime")
@get:NotNull
public val createdDateTime: LocalDateTime,
@param:JsonProperty("createdInstant")
@get:JsonProperty("createdInstant")
@get:NotNull
public val createdInstant: Instant,
...
```

F43nd1r:
We'd also like to parse string fields into types, but based on custom `format`

PR #685, opened 2026-09-14
Add custom type mapping

F43nd1r:
Allows to specify custom type mappings using the format `type:format=KotlinFqcn[;kotlinx=SerializerFqcn]`

Maps from openapi type+format to class name and optionally serializer.

Related issue: #639.

Example:
```yaml
components:
  schemas:
    Content:
      type: object
      required:
        - id
      properties:
        id:
          type: string
          format: duration
```
with 
```
--custom-type-mapping "string:duration=java.time.Duration;kotlinx=com.example.DurationAsIsoStringSerializer"
```

maps to 
```kotlin
@Serializable
public data class Content(
  @Serializable(with = DurationAsIsoStringSerializer::class)
  @SerialName("id")
  @get:NotNull
  public val id: Duration,
)
```


cjbooms:
@ulrikandersen any opinion here? I am all for flexibility, but worried that this adds an additional control surface that could be confusing.

ulrikandersen:
I’m also in favour of flexibility, but I share the concern about adding another configuration surface that we’ll need to maintain over time. Could this be handled on the consumer side instead; for example, keeping the generated property as a string and converting it to the desired type in a wrapper or mapping layer? 

That does mean some extra consumer code, but keeps application-specific types and serialization choices outside the generator. 

Would that work for your use case?

F43nd1r:
While certainly possible to do this in a post-processing step, it loses a bit of convenience:
 - type information from the schema is "lost" - a developer needs to know/check what a string is actually supposed to represent
 - two-phase serialization means also two-step error handling - an instant leads to a deserialization failure in the first step, while a duration leads to a parsing failure in the second step
 - we have one custom type (Ulid), which is used in hundreds of places, in the generator this is one config but in postprocessing we'd have to touch everything.

Background info: We're looking to migrate from openapi-generator, because it has issues with polymorphism. My contributions here in this repo serve to bridge the feature gaps we identified (in this case https://openapi-generator.tech/docs/usage/#type-mappings-and-import-mappings). 
We're hoping to gain the better polymorphism support of Fabrikt while not making steps back in other places.

ulrikandersen:
Thanks for the additional context. Looking more closely at the code, the changes seem fairly contained and fit well with the existing extension points. That helps address my concern about the impact on the core logic, and I’m more comfortable with this approach now.

Thanks for the contribution 🙏

ulrikandersen (approved):

PR #698, opened 2026-09-16
Add a configuration option to allow operationIds to be transformed

### LINKED ISSUE #682: Remove redundant information from client method name

cjbooms:
Closes #682
Supersedes #684 

Adds an option to allow operationId to be transformed before it is used in function names in Controller interfaces and Clients.


ulrikandersen (approved):
Nice to see a more generic solution to the problem 👍

F43nd1r:
Yeah this looks good :+1: 

only edge case that could break is if someone tried to use a regex containing `:`

cjbooms:
> Yeah this looks good 👍
> 
> only edge case that could break is if someone tried to use a regex containing `:`

Good point, but acceptable I would say

Issue #702, opened 2026-09-16
ModelNameRegistry collision detection doesn't fire when a cache hit satisfies getOrRegister before register() runs


cjbooms:
## Summary

An inline enum property and an unrelated top-level component schema can compute the same generated class name string, and `ModelNameRegistry` detects this collision only when both callers reach `register()`. If the first caller's lookup is satisfied by a cache hit (`this[tag]`) rather than a fresh `register()` call, the second, distinct schema silently receives the first schema's already-allocated name instead of a de-duplicated one. The generated model then references a class with the wrong shape (a data class where an enum was required, in the reproduction below), and the output does not compile.

This is different from #622 (bounded `Extra` suffix growth): that issue is about the suffix growing unboundedly *when collision detection does fire*. This issue is about collision detection *not firing at all* for one real shape.

## Root cause

`ModelNameRegistry.resolveTag(schema, modelClassName)` builds its cache key as

```kotlin
"file:${uri.file}#$modelClassName"
```

— a function of the *computed name string* only, never of the schema's own identity (e.g. `schema.jsonReference`). Two structurally unrelated schemas that happen to compute the same `modelClassName` therefore share one tag.

Concretely, given:

```yaml
components:
  schemas:
    Product:
      type: object
      properties:
        state: { type: string, enum: [live, draft] }
    ProductState:
      type: object
      properties:
        state:
          $ref: "#/components/schemas/Product/properties/state"
```

The top-level `ProductState` component names itself `"ProductState"` via `schema.name`. `Product.state`'s inline enum names itself via `enclosingSchema("Product").toModelClassName() + safeName("state").toModelClassName()` = `"Product" + "State"` = `"ProductState"` — the same string, by pure coincidence of naming convention, not because the two schemas are related.

Whichever of the two is processed first calls `register()` and populates `tagToName["file:...#ProductState"] = "ProductState"`. Whichever is processed second calls `getOrRegister()`, which checks `this[tag]` *before* ever calling `register()` (see `getOrRegister`'s `.getOrElse { register(...) }` pattern) — that lookup succeeds, so the second schema is returned the *first* schema's name without ever exercising the collision-detection branch inside `register()` (the `IllegalArgumentException` on a duplicate tag is only reachable from a second `register()` call, never from a second `getOrRegister()` call that hits the cache).

[...]

PR #704, opened 2026-09-16
Fix inline enum inside a map value never generating a backing class

cjbooms:
An additionalProperties value schema's own inline properties were passed to standardDataClass directly, never through buildInLinedModels, so an inline enum nested inside a map value never got its backing class emitted -- the generated code referenced a class that was never written.

Both buildMapModel and the AdditionalProperties branch in buildInLinedModels now recurse into buildInLinedModels for their computed properties, mirroring ObjectInlinedField's existing pattern.

Reproduced independently of #702/#703 (confirmed byte-identical broken output on a pre-fix checkout). Added a regression schema to mapExamples. Verified against the real Scayle Admin API spec: the generated output now compiles end to end where it previously failed on an unresolved reference in a webhook targetSchema definition.


cjbooms:
Superseded by a clean branch cut directly from master instead of stacked on the #703 branch: #705

PR #708, opened 2026-09-17
Generate models from JSON Schema documents, not just OpenAPI 3

### LINKED ISSUE #707: Generate models from JSON Schema documents, not just OpenAPI 3

**Problem:** fabrikt only accepts OpenAPI 3 documents, but a large amount of real schema material is plain JSON Schema nested inside another resource — Nakadi `EventType` manifests, Confluent Schema Registry subjects, AsyncAPI payloads, Kubernetes CRDs. Today those need a bespoke conversion script per project before fabrikt can be used at all.

**Proposal:** a parser-agnostic precursor step that converts JSON Schema (draft-04 through 2020-12) into an OpenAPI 3.1 document. It's opt-in via a trailing `#/json/pointer` fragment on `--api-file` (RFC 6901, the same syntax already used in `$ref`), plus `--json-schema-root-name` for when the schema has no `title` and the resource no `/metadata/name`. The fragment addresses the schema within a larger resource, e.g. `manifest.yaml#/spec/schemaObject` for a Nakadi `EventType` manifest, so one mechanism serves every wrapper shape without a dedicated flag per ecosystem.

**Scope:** Kubernetes CRD support is *not* included in this proposal, though the generic pointer mechanism is expected to be reusable for it later. Unconvertible keywords (`if`/`then`/`else`, `patternProperties`, `contains`, `propertyNames`, `dependencies`, `unevaluated*`) are dropped rather than approximated — they have no OpenAPI or fabrikt equivalent, so their constraints would not be enforced in generated code either way.

**Why in fabrikt rather than each caller:** the conversion is dialect knowledge, not project knowledge, and every caller reimplementing it produces subtly different models from the same schema.

### Example Spec

```yaml
apiVersion: nakadi.zalan.do/v1
kind: EventType
metadata:
  name: pop.offers-config-test
spec:
  owningApplication: pop-offers
  category: data
  schemaObject:
    $schema: 'http://json-schema.org/draft-04/schema#'
    definitions:
      Money:
        type: object
        properties:
          amount:
            type: number
          currency:
            type: string
        required: [amount, currency]
    properties:
      total:
        $ref: '#/definitions/Money'
    required: [total]
```

Invocation:

```
fabrikt --api-file 'event-type.yaml#/spec/schemaObject' \
  --json-schema-root-name OffersConfig \
  --base-package com.example.offersconfig \
  --targets http_models
```

### Desired Output

```kotlin
data class Money(
    val amount: BigDecimal,
    val currency: String,
)

data class OffersConfig(
    val total: Money,
)
```

cjbooms:
Fabrikt can now generate models from a plain JSON Schema document (draft-04 through 2020-12), not just a full OpenAPI 3 spec. This closes #707.

### How to enable it

`--json-schema-file` takes a JSON Schema document as the primary input instead of `--api-file`, converting it to an OpenAPI 3.1 document before generation runs. A trailing `#/json/pointer` fragment (RFC 6901 — the same syntax `$ref` already uses) selects the schema nested inside a larger resource. No new dependency, no separate conversion step: point `--json-schema-file` at the manifest and generate as normal.

Given a Nakadi `EventType` manifest — a JSON Schema document nested under `spec.schemaObject`, one of the motivating real-world cases:

```yaml
# manifest.yaml
apiVersion: nakadi.zalan.do/v1
kind: EventType
metadata:
  name: pop.offers-config-test
spec:
  owningApplication: pop-offers
  category: data
  schemaObject:
    $schema: 'http://json-schema.org/draft-04/schema#'
    definitions:
      Money:
        type: object
        properties:
          amount:
            type: number
          currency:
            type: string
        required: [amount, currency]
      State:
        type: string
        examples: [AVAILABLE, UNAVAILABLE]
    properties:
      total:
        $ref: '#/definitions/Money'
      state:
        $ref: '#/definitions/State'
    required: [total, state]
```

Run:

```
java -jar fabrikt.jar \
    --json-schema-file 'manifest.yaml#/spec/schemaObject' \
    --json-schema-root-name 'OffersConfig' \
    --base-package 'com.example' \
    --targets 'http_models' \
    --http-model-opts 'x_extensible_enums'
```

`--json-schema-root-name` names the schema built from the pointed-at schema's own top-level `properties`; it falls back to the schema's `title`, then the resource's `/metadata/name`, then a generic default with a warning if neither is present — here it's needed because the manifest has neither.

This generates `Money.kt`, `State.kt`, and `OffersConfig.kt`. `State`'s top-level `{type: string, examples: [...]}` is a de-facto enum, so with `x_extensible_enums` it becomes:

```kotlin
public enum class State(
  @JsonValue
  public val `value`: String,
) {
  AVAILABLE("AVAILABLE"),
  UNAVAILABLE("UNAVAILABLE"),
  ;

  override fun toString(): String = value

  public companion object {
    private val mapping: Map<String, State> = entries.associateBy(State::value)

    public fun fromValue(`value`: String): State? = mapping[value]
  }
}
```

`OffersConfig` references `Money` and `State` exactly as `$ref`s to a normal OpenAPI component would. `--api-fragment` still works alongside the pointer, so common component schemas can be shared across multiple converted manifests.

See the README's "Generating from a JSON Schema document" section for the full conversion behavior (dialect coverage, dropped keywords, `$defs` handling, exclusive-bound upgrading).

### Scope

Kubernetes CRD conversion is **not** a supported target — the vendored CRD fixture (`src/test/resources/examples/jsonSchemaConversion/dialects/`) only asserts the converter doesn't throw; no CRD-specific behavior is implemented or tested.

The one behavioral change to existing (non-JSON-Schema) generation is gated: numeric exclusive bounds are downgraded to the boolean form `PropertyUtils` already reads, but only for documents declaring `openapi` above 3.0 — no existing 3.0 example's output changes.


cjbooms:
Anyone have any thoughts about this # based approach to config?
```
    --api-file 'manifest.yaml#/spec/schemaObject' 
```

The alternative is a completely separate CLI flag for input, but I think ^^ is cleaner

mrhoeve:
I like the `#/spec/schemaObject` part for selecting a nested schema — it is compact, familiar from `$ref`, and works equally well for local files and URLs.

I am slightly less convinced about using the presence of the fragment to also select JSON Schema mode. It means a standalone root schema has to be written as `schema.json#`, while any existing `--api-file` value containing `#` changes meaning, including valid local filenames and URL fragments.

I would lean towards a mutually exclusive `--json-schema-file 'manifest.yaml#/spec/schemaObject'`. That keeps the fragment as the selector, but makes the input format explicit and preserves the existing meaning of `--api-file`. It also seems like a clearer boundary if Fabrikt’s JSON Schema support grows beyond this initial conversion step.

The current approach is still workable if keeping the flag count down is the stronger priority; I do not think it should block the PR. My preference for the separate flag is mainly about CLI compatibility and making the mode switch unsurprising.

cjbooms:
> I would lean towards a mutually exclusive `--json-schema-file 'manifest.yaml#/spec/schemaObject'`. That keeps the fragment as the selector, but makes the input format explicit and preserves the existing meaning of `--api-file`. It also seems like a clearer boundary if Fabrikt’s JSON Schema support grows beyond this initial conversion step.

Thanks for the feedback. Agreed, Let's go with that. Will make the change on Monday

PR #713, opened 2026-09-21
Generate models for inline operation objects

### LINKED ISSUE #187: Support generation of models from operation-level properties
At the moment, Fabrikt only supports generation of models from global schemas. Non-simple schema types that are defined "inline" e.g. in operation parameters are not generated. Similar to #20 , this can probably be done by extending `SourceApi.init`.

mrhoeve:
## Summary

Fabrikt generates models for schemas under `components`, but inline object schemas defined directly on operations do not consistently become named Kotlin models. As a result, generated parameters, request bodies, and successful response bodies can lose their concrete type, while inline error-response models are missing entirely.

This PR discovers and names inline object schemas in operation parameters, non-multipart request bodies, successful responses, and 4xx, 5xx, and `default` error responses. It also generates models for inline object items in response arrays. For example, an operation with inline `200 → Widget` and `404 → Problem` bodies now generates models for both, just as it would if those schemas were defined under `components/schemas`.

## Naming and scope

An explicit schema `title` supplies the model name. Otherwise, Fabrikt derives a stable name from the operation; error-response names also include the response status so distinct errors remain distinguishable. When an operation has no `operationId`, the HTTP method and path provide the fallback name.

The change covers direct objects and `allOf` object shapes. For successful responses and non-multipart request bodies, Fabrikt generates an operation-level model when there is a single distinct inline schema to name. Existing handling of `oneOf`, `anyOf`, multipart bodies, and component schemas remains unchanged.

## Generated behavior

Generated clients and controllers use the named models for supported inline inputs and successful responses instead of falling back to `Any`. Inline error-response models are generated and available to consumers; this PR does not change how clients or controllers handle non-success responses. Specs outside the supported inline shapes retain their existing generated output.

Fixes #187.


ulrikandersen:
Thanks for working on this! Awesome that inline objects now get concrete model names. I've been using a separate pre-processing step for this - not so nice 😁 

One gap I’d like to discuss is inline error responses. Today, if an operation returns 200 → Widget and 404 → Problem, with both schemas defined under components/schemas, we generate both models. With this PR, equivalent inline definitions would only generate Widget. Could we discover and name the inline error schemas too?

Could we also add coverage across all controller and client targets? Just to make sure we lock down the behaviour across the board.

mrhoeve:
Thanks for pointing this out, Ulrik. You were right: the PR generated a model for an inline 200 response but skipped an equivalent inline 404 response.

I’ve added model discovery for inline 4xx, 5xx, and `default` responses, including object items in response arrays. Without a schema title, the generated name includes the operation and response status. I’ve also added coverage across all client and controller targets for OpenAPI 3.0, 3.1, and 3.2.

The follow-up commit is now in the PR, and I’ve updated the PR description to reflect the expanded scope and clarify that client and controller error handling remains unchanged. The full build passes. Thanks for catching the gap!

ulrikandersen (approved):
Thanks for adressing that, @mrhoeve. 
Using the status code as part of the model name works well IMO.

PR #714, opened 2026-09-21
Generate `@Deprecated` annotations from OpenAPI metadata

mrhoeve:
## Description

OpenAPI schemas can mark schemas, properties, operations, and parameters as deprecated, but Fabrikt currently discards that metadata during code generation.

This change preserves the deprecation metadata through the parser facade and generates Kotlin `@Deprecated` annotations for:

- schemas and enum types;
- model properties;
- API operations;
- operation parameters.

The behavior is supported by all client and controller generators. The generated messages are centralized so every target produces consistent annotations.

Tests cover OpenAPI 3.0.4, 3.1.2, and 3.2.0, including deprecated and active elements in the same specification. Existing golden examples remain unchanged.

## Validation

- `./gradlew clean build`
- All unit, golden-file, playground, and end-to-end tests pass.

Advances #656.


ulrikandersen (changes requested):
Thanks for this! I like the direction and the coverage across all client and controller targets.

One blocker before merging: Kotlin’s [`@Deprecated`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin/-deprecated/) isn’t applicable to function parameters AFAIK, so the generated `findSubject(@Deprecated(...) id: String)` will fail compilation.

Could we use KDoc for parameter deprecation instead, keeping annotations on models, properties, and operations?

With that addressed, I’d be happy to accept this.

mrhoeve:
Thanks for catching this, Ulrik. You’re right: I incorrectly generated Kotlin’s `@Deprecated` on function parameters, where it isn’t allowed. I missed it because the test inspected the generated code without compiling that specific example.

I’ve removed the parameter annotation and now document `deprecated: true` in the parameter’s `@param` KDoc across the client and controller generators. The annotations on models, properties, and operations remain. I’ve updated the regression checks, and the full build passes. Thanks for spotting the mistake so precisely.

ulrikandersen (approved):

PR #715, opened 2026-09-22
Generated annotation

### LINKED ISSUE #674: Ensure Java / Kotlin code is marked as generated

Any generated class must be annotated with `jakarta.annotation.Generated` (or similar, such as `javax.annotation.processing.Generated`) to simplify excluding the code from various quality tools. Some quality tools allow exclusion by filesystem path or package; others require annotations to exclude the code. This change will improve ease of use for JVM-based consumers.

By adding the annotation, generated code can easily be excluded from quality analysis by standard JVM tooling.

### Example Spec

```yaml
openapi: 3.0.0

info:
  title: Pets API
  version: 1.0.0

paths:
  '/pets':
    get:
      operationId: listPets
      summary: List all pets
      responses:
        '200':
          description: A list of pets
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/Pet'
    post:
      operationId: createPet
      summary: Create a pet
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
      responses:
        '201':
          description: Pet created
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'

  '/pets/{petId}':
    get:
      operationId: getPet
      summary: Get a pet by ID
      parameters:
        - name: petId
          in: path
          required: true
          schema:
            type: string
      responses:
        '200':
          description: A single pet
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/Pet'
        '404':
          description: Pet not found

components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
        - type
      properties:
        id:
          type: string
        name:
          type: string
        type:
          type: string
          enum:
            - dog
            - cat
        age:
          type: integer
```

### Desired Output

```kotlin
// Pet
//
// This file was generated from an OpenAPI specification by Fabrikt.
// DO NOT EDIT. Any changes will be overwritten the next time the code is generated.
// To update, modify the specification and re-generate.
//
package com.example.models

import jakarta.annotation.Generated
import jakarta.validation.constraints.NotNull
import kotlin.Int
import kotlin.String
import kotlinx.serialization.SerialName
import kotlinx.serialization.Serializable

@Generated(value = "io.fabrikt:fabrikt", date = "2001-07-04T12:08:56.235-0700", comments = "Generated with Fabrikt v27.0.1")
@Serializable
public data class Pet(
  @SerialName("id")
  @get:NotNull
  public val id: String,
  @SerialName("name")
  @get:NotNull
  public val name: String,
  @SerialName("type")
  @get:NotNull
  public val type: PetType,
  @SerialName("age")
  public val age: Int? = null,
)
// PetType
//
// This file was generated from an OpenAPI specification by Fabrikt.
// DO NOT EDIT. Any changes will be overwritten the next time the code is generated.
// To update, modify the specification and re-generate.
//
package com.example.models

import jakarta.annotation.Generated
import kotlin.String
import kotlin.collections.Map
import kotlinx.serialization.SerialName

@Generated(value = "io.fabrikt:fabrikt", date = "2001-07-04T12:08:56.235-0700", comments = "Generated with Fabrikt v27.0.1")
public enum class PetType(
  public val `value`: String,
) {
  @SerialName("dog")
  DOG("dog"),
  @SerialName("cat")
  CAT("cat"),
  ;

  override fun toString(): String = value

  public companion object {
    private val mapping: Map<String, PetType> = entries.associateBy(PetType::value)

    public fun fromValue(`value`: String): PetType? = mapping[value]
  }
}
```

ulrikandersen:
This pull request introduces a new feature to support annotating generated Kotlin types with `@Generated` from `javax.annotation.processing` controlled by a new output option:

```
@Generated(
    value = ["io.fabrikt.cli.CodeGen"],
    comments = "Generated with Fabrikt v27.0.1",
)
```

Plays well together with the existing `ADD_FILE_DISCLAIMER` option to clearly mark generated code as generated.

The PR adds golden files for every controller & client target in addition to the models in order to lock down the functionality across the generation surface.

Closes #674


mrhoeve:
Thanks, Ulrik — this fits well alongside the direction in #656. I see it as a focused improvement to generated-code metadata rather than a step in the OpenAPI 3.1/3.2 parsing work. Keeping it opt-in preserves existing output, and covering the client and controller targets matches the incremental approach we discussed.

One question before merging: `GenerationMetadata` defaults to `Instant.now()`, and the date is emitted whenever the option is enabled. Rerunning Fabrikt on an unchanged spec would therefore change the generated files. Could we omit the date unless an explicit, reproducible value is supplied? That would keep the generator and version metadata useful without creating unnecessary diffs.

ulrikandersen:
Yes that's a good point. If generated files are checked in to version control the diff would show regenerated files even though the content is the same. We could also choose to simply omit the date. I would be perfectly fine with that solution.

PR #716, opened 2026-09-22
Handle OpenAPI 3.1 type arrays safely

mrhoeve:
Description:

## Summary

- derive OpenAPI 3.1 type-array semantics through the existing schema facade for component, referenced, and operation-level schemas
- preserve a single non-null type combined with `null` as its concrete nullable Kotlin type
- generate `Any` for Jackson or `JsonElement` for kotlinx.serialization when a schema declares multiple non-null types, with a warning that explains the fallback
- preserve existing generated output outside the newly supported type-array cases

## Why

OpenAPI 3.1 inherits JSON Schema type arrays, but the existing Kaizen representation cannot preserve every declared type. Selecting one member of a heterogeneous type array would produce an unsafe Kotlin contract. Routing the native parser semantics through the established facade allows Fabrikt to retain precise nullable types where possible and otherwise generate a compileable, explicit fallback without exposing parser internals to generation code.

## Generated output impact

Schemas with one concrete type plus `null` continue to generate the corresponding nullable Kotlin type. Newly covered schemas with multiple non-null types generate `Any` with Jackson or `JsonElement` with kotlinx.serialization instead of claiming one unsupported concrete type. Existing golden output remains unchanged.

Advances #656


mrhoeve:
I also reviewed the README for this change. It now explains how OpenAPI 3.1 type arrays map to Kotlin, including nullable single types and the Jackson and kotlinx.serialization fallbacks for multiple non-null types.

I updated the older inline-schema guidance as well, so it reflects the support added in #713. The two README updates are separate commits in this PR.

